PR #9, opened 2024-05-31
added console command to replicate benchmark results

daiwaid:


benchmark/README.md
```diff
@@ -5,6 +5,11 @@
 | :------- | :----- | :----------------------------------   |
 | `gpt-4o` | Off    | Schema inline with entity description |
 
+### Replicate
+```
+python -m benchmark.run --dataset <dataset-name> --num_pairs 0
```
zenodflow, line 10: Is it possible to make the --num-pairs "-1" instead? @daiwaid

zenodflow: .. for "all pairs", assuming that's what "0" means

daiwaid: actually I had it so any number <= 0 meant full dataset so I just had 0 since that's 1 less keystroke to type

zenodflow (approved):
Lgtm

PR #10, opened 2024-06-01
feat: add duckduckgo search and make it default

daiwaid:
simple duckduckgo search using the langchain tool, some refactoring to separate search-engine-specific code


zenodflow (commented):
Looks good. @daiwaid Did you test the examples/browse.py and see if it still works as expected?

zenodflow:
If browse still works: https://github.com/abcsys/libem/blob/main/examples/browse.py, let's make duckduckgo the default?

daiwaid:
browse.py works and duckduckgo is the default.

zenodflow:
Nice!

zenodflow (approved):

PR #14, opened 2024-06-04
feat: improvements to the browsing tool

daiwaid:


libem/match/prompt.py
```diff
@@ -16,7 +16,7 @@
 )
 
 output = Prompt(
-    default="Answer with 'Yes' if they do and 'No' if they do not.",
+    default="Answer with a single 'Yes' if they do and 'No' if they do not.",
```
zenodflow, line 19: Is there a reason why "a single" is needed?

libem/browse/engine/duckduckgo.py
```diff
@@ -1,8 +1,22 @@
...
+from duckduckgo_search.exceptions import TimeoutException
 
 def search_wrapper(k=1, full_text=False):
     wrapper = DuckDuckGoSearchAPIWrapper(max_results=k)
-    search = DuckDuckGoSearchRun(api_wrapper=wrapper)
+    engine = DuckDuckGoSearchRun(api_wrapper=wrapper)
+    
+    def search(query):
+        tries, result = 0, None
+        while result is None and tries < 3:
```
zenodflow, line 11: Parameterize "3" in parameter.py as num_retry?

zenodflow (commented):
I pulled the branch and tested it. An observation: sometimes Libem would trigger "Warning: Max call reached", I wonder if it is because the tool use call returns new tool calls?

libem/match/prompt.py
daiwaid: yes otherwise when it uses tools it starts to give explanations for the answer

libem/browse/function.py
```diff
@@ -15,18 +15,18 @@
...
+                "query": {
                     "type": "string",
-                    "description": "Entity description",
-                },
-                "engine": {
-                    "type": "string",
-                    "description": "The search engine to use. Choose from "
-                                    "['google', 'duckduckgo'], default is "
-                                    "'duckduckgo'.",
+                    "description": "Query to search.",
```
zenodflow, line 20: If we do not let the model choose which engine to use, should we then make the engine a parameter?

daiwaid: yeah probably, I commented it out because it would choose google sometimes and we don't want it to choose google right since google is not free

zenodflow:
Closing for now - wait for browse tool design.

PR #17, opened 2024-06-05
feat: make the options of Parameter selectable via an Index API

daiwaid:


libem/browse/parameter.py
```diff
@@ -1,9 +1,9 @@
-from libem.core.struct import Parameter
+from libem.core.struct import Option, Parameter
 import libem.parameter as libem
 
 model = libem.model.copy()
 temperature = libem.temperature.copy()
 engine = Parameter(
-    default="duckduckgo",
-    options=["google", "duckduckgo"],
+    default=1,
```
zenodflow, line 7: What is "1"?

zenodflow (commented):
Could you explain how the "option" field is used?

libem/browse/parameter.py
daiwaid: option is a list and you can pass an index to a Parameter variable to change the current value to a value stored in options

daiwaid: And when initializing, the 'default' field can also be just an index into 'options', which will set the current value to that value stored in 'options'.

zenodflow:
I think the semantics of "default" after this change is more akin to "current_index" rather than the intended semantics of "default_value"?

Either case, who will configure this index?

The original goal of the "options" is that tool developers can add options to parameters and prompts so that the tuning or optimization process can use thee {default} U {options} as hints or search space to figure the best values.

daiwaid:
I think default still means default here because that only indicates the initial value the parameter takes on (in the parameter.py files). Then after that, one can use calibrate() to select a different option from the 'options' list or set the parameter to an entirely new value.
The new semantics still allow for the {default} U {options} search space, except the default value is part of options now, so there is no longer a need for the union.

zenodflow:
I see your point. It does make the file a bit hard to read though (e.g., one needs to parse the default as an index). To confirm:
- Could we still make the default a value and not an index and when it is an index, it can point to the options? I think the API would be clearer as `default=index(1), options=[...]`.
- Do we have to make the values inside options an Option type? Could we make the type optional? I.e., python type still allowed? Or, if all values inside the options are of type Option, could we just do a map() or type cast when we load the options? What is the purpose for the Option type?

daiwaid:
Yes, that would work. The reason why we need an extra class (Option) is to distinguish index vs value. We could change it so that when we index we wrap it in an Index class instead, then the Option class is no longer necessary. As for letting default still assignable to a value when options field is not empty, it’s fine but I think all options should still appear as part of the options list so it’s easy to tell what all the options are at a glance (as well as no ambiguity for indexing).

libem/core/struct.py
```diff
@@ -13,14 +13,32 @@ def search(self, train_data, metric):
...
     def __init__(self,
-                 default: typing.Any,
-                 options: list[typing.Any] = None
+                 default: Index | typing.Any = 0,
+                 options: list[typing.Any] = []
                  ):
-        self.value = self.v = default
+        if isinstance(default, Index):
+            assert len(options) > default()
+        else: # quick assign of a single value: set default to value
```
zenodflow, line 32: What does "quick" mean? Maybe just "use default as the only available option"

libem/core/struct.py
```diff
@@ -13,14 +13,32 @@ def search(self, train_data, metric):
...
+            if len(options) > 0:
+                raise TypeError("Either set 'default' to an Index into the options list "
+                                "or leave options empty to quick assign a single value.")
```
zenodflow, line 35: This message should be "The default should be an Index when options are present."

libem/match/parameter.py
```diff
@@ -1,18 +1,19 @@
...
-from libem.core.struct import Parameter
+from libem.core.struct import Index, Parameter
 
 model = libem.model
 temperature = libem.temperature
 
 tools = Parameter(
-    default=[],
-    options=[],
+    default=Index(1),
```
zenodflow, line 8: Index(0)

libem/match/prompt.py
```diff
@@ -1,10 +1,10 @@
-from libem.core.struct import Prompt, Shot
+from libem.core.struct import Index, Prompt, Shot
 
 """System prompts"""
 role = Prompt(
-    default="You are an entity matcher that determines whether "
-            "two entity descriptions refer to the same real-world entity.",
-    options=[""],
+    default=Index(0),
```
zenodflow, line 5: I don't think Libem knows what task to perform if it's given ""

zenodflow (changes requested):
Left some comments. 

I am still unsure how to use the Index() API though. If we run optimization or parameter / hyper search, then the value will be set from the user with libem.calibrate, so there is no need of an index per se. 

I think the goal of "options" and "default" are simply for developer to specify possible values and default value, and there is no "state tracking" for which value to use. The actual value is always set by value rather by index (unless there is a reason that I miss).

How do you see users to use the Index() API?

daiwaid:
The Index method is meant to be a much faster way of setting the current value of a parameter to a different pre-defined option rather than copy-pasting the entire option over, which could be multiple lines. It also helps with not forgetting to update the options list when tweaking an option in a separate file (e.g. experimenting with different wordings for a prompt).

zenodflow:
> much faster way of setting the current value of a parameter 

So the idea is to enable someone to `libem.calibrate({"libem.match.parameter.tools": [Index[2]]})`?

daiwaid:
> parameter.py and prompt.py are never meant for _users_ to look at or configure
If that's the case then how does a user know what parameters are available?

And for the proposed `get().options()` route, it's 1. more code to write each time and 2. slightly more resource-heavy (need to deep copy the options list to avoid accidentally mutating it somewhere and causing indexing to result in unexpected behavior).
Since we already have a non-standard way of setting these parameter values (`libem.calibrate("libem.match.parameter.cot": True)`), why not have the option to index into the options list directly (`libem.calibrate("libem.match.parameter.cot": Index(1))`) rather than doing `libem.calibrate("libem.match.parameter.cot": libem.calibrate.get("libem.match.parameter.cot").options()[1])`?

This PR is mainly for developers like us to be able to code and test new features (like new benchmark files) efficiently and avoid introducing bugs like inconsistent values listed in the options list vs what it is being changed to in the new file.

zenodflow:
> This PR is mainly for developers like us to be able to code and test new features (like new benchmark files) efficiently and avoid introducing bugs like inconsistent values listed in the options list vs what it is being changed to in the new file and not for the eventual end users.

I see, the proposed goal is to introduce a syntactic sugar for Libem internal developers? Could you explain how does it help code new features efficiently?

Re. Introducing bugs: does the current approach have inconsistent value bugs?

zenodflow:
> libem.calibrate("libem.match.parameter.cot": libem.calibrate.get("libem.match.parameter.cot").options()[1])

Why does a _user_ would use `[1]` to index into the options in this example?

I believe one possible workflow for using the options is, continuing the previous message:
- `p = libem.calibrate.get("libem.match.parameter.tools")`
- `p_opt = search(p.options())`
- `libem.calibrate({"libem.match.parameter.tools": p_opt})

The above does not involve an index. Basically, we need a concrete use case of the Index API to justify the change of current interface, and it's still unclear to me what that is. I'm not sure fixing the inconsistent use of `default` field is a feature as that is introduced by the Index() API itself. For the "writing less code" point, I want to see the client code first -- and not that the code is shorter, but it is _clearer_ -- unless there is performance implications.

daiwaid:
> Could you explain how does it help code new features efficiently?

Yes, this PR is essentially syntactic sugar for developers. The best use case for this currently is when switching between long prompts. For example, suppose we add domain-specific prompts as options under `libem.match.prompt.query` like this:
```
query = Prompt(
    # default=...,
    options=["Entity 1: {left}.\nEntity 2: {right}.", 

             "Do the two publications refer to the same real-world publication? "
             "Publication 1: '{left}'\nPublication 2: '{right}'",

             "Do the two song descriptions refer to the same real-world song? "
             "Song 1: '{left}'\nSong 2: '{right}'",

             # ...
             ],
)
```
And if in `benchmark/classic/itunes_amazon.py` I want to use the song prompt, then currently I have to copy-paste the 3 lines into that file. Then suppose I do prompt engineering on it while testing `itunes_amazon.py` and find that this prompt performs better
```
             "Are the following two two song descriptions referring to the same song? "
             "Song A: '{left}'\nSong B: '{right}'"
```
Then I have to remember to copy-paste this new prompt back into `libem/match/prompt.py` every time. And what if there is another file in `benchmark/???/itunes_amazon.py` that also uses the song prompt? Then I will need to update that file as well.

> `p_opt = optimize.search(p.options())`

This means we have to re-run the optimize algorithm for that parameter every time someone wants to run the file, which can add unnecessary latency. Take the above `itunes_amazon.py` benchmark example, since we already know the domain of the dataset, we don't need libem to figure out the best domain prompt every time someone wants to run the benchmark.

zenodflow:
I see. Making the right options available to be picked at runtime is indeed a right use case for libem! However, I think we need more than a syntactic sugar to provide this feature. In the example you gave above, one possible client code should be doing similarity match between some dataset metadata and the options and pick the most similar one.
- The proposed API assumes an organization (i.e. list order) of the options and uses a static index to retrieve the parameter option. This introduces a dependency between the client code (e.g., your benchmark script) and the tool code (e.g., match's prompt.py). This means one / the developer needs to maintain this "long-distance" dependency, e.g., when new options are introduced.
- Because such dependency can exist, I think having the client code (e.g., your benchmark script) to explicitly _figure out_ (e.g., similarity, copy-over) which option to use makes the code clearer -- we don't need to worry about the dependency. 

The problem you encountered seems to be addressable by using `libem.calibrate({"libem.match.prompt.role": PROMPT})`. In fact, that is the right way it should be done -- the tool's prompt should only be updated by the tool developer, not the user / the client code / your benchmark script. 

For clarity, in Libem, a parameter should have the following types of values:
- Current value (set to default value at initialization time if not overwritten later at runtime by client code)
- Default value 
- Option values
One uses `libem.calibrate()` to set the current value and to use `libem.calibrate.get()` to retrieve the domain / all options of the parameter (union of default and options). This is just what I had in mind and open to discussion / alternatives!

> This means we have to re-run the optimize algorithm for that parameter every time someone wants to run the file, which can add unnecessary latency. 

No. The optimization workflow figures out the parameter and stores it somewhere (could be the catalogue or parameter.opt). To clarity, what I meant was the optimization workflow is so far I can imagine how a client code would use the options. I don't think any client code would involve it as part of the calibration process. They can load the optimal parameters from catalogue though.

daiwaid:
> having the client code (e.g., your benchmark script) to explicitly figure out (e.g., similarity, copy-over) which option to use makes the code clearer

The current approach (copy-over) still runs into the issue of if the tool developer modifies the prompts, all the client code will not automatically update to the new prompt. I agree that indexing isn't perfect but I can't think of something else that's strictly better. Here's my take on the plausible solutions: 

indexing (list):
- pro: least amount of code (just an index)
- con: existing options can be modified but cannot be removed and new options need to append to the end

key-value pairs (dict):
- pro: order can change, relatively short code
- con: existing options can be modified but cannot be renamed/removed, need to come up with a key naming convention, options order not (strictly) guaranteed (bad if trying to optimize over a range)

string similarity:
- pro: resilient to small modifications
- con: unpredictable behavior with how the similarity function will score each option (a new option could get a better score than the desired one, a modified option could get a lower score than other existing options), possibly more code than copy-paste

I think it is unlikely that an existing option will get removed, and there is little incentive to re-order the options (maybe if the options is a range and that range changes?), but modifying existing values and adding new values can reasonably occur. With this, I think indexing is still the best solution. But feel free to suggest other options if you think of any.

> The problem you encountered seems to be addressable by using `libem.calibrate({"libem.match.prompt.role": PROMPT})`.

This does not solve the problem of modifying an existing option somewhere and ensuring that the change propagates to everywhere else that uses said option.

> the tool's prompt should only be updated by the tool developer, not the user / the client code / your benchmark script.

Shouldn't the tool developer be the one to provide the benchmark script to benchmark the tool? (or integrate the tool as part of the existing benchmark)? I'm not really following your intuition for the strict separation between a libem internal module developer vs a downstream client developer/end user. Doesn't the open-source nature mean that any end user can also become an internal developer if they need a functionality that the library currently doesn't support?

zenodflow:
@daiwaid Nice breaking down of the issue and comparing the options.

I think usually keeping things simple (but not naive) is the way to go if the addition of a feature has both pros and cons.

> indexing (list):

I'd say the main con of this is that it introduces a dependency across client code and tool code. It makes the client code rely on assumptions about the tool code (or vice-versa). 

> key-value pairs (dict):

Same as before. Client code needs to know the key of the value. Also unclear what key to use. However, this approach would make sense if there is a canonical way of naming those prompt options (e.g., one that is tool specific!).

> string similarity

This one should be used as a calibration strategy and only meant for an optimization (hence it does not need to be exact -- and the developer of the client code is assumed to be fully aware of it). I wouldn't say it is comparable to the previous two (i.e., optimization vs. referencing).

Re. copy-over: I think there is a misunderstanding of what I meant by copying over. What I meant was: if the developer of the client code needs some options, she copies those option _values_ over what's in the tool's options. There is no explicit coupling of those values with the tool values. Who read the client code knows _exactly_ what values are used.  
- Note that the _implementation_ of copy-over does _not have to_ be manually copying over code (also, less code does not always mean better code / modularity), they can use the parameter.options() to get them.
- However, I would argue, if it were not for optimization, copying over the values as code is the right way to go.
In other words, what you pointed out "all the client code will not automatically update to the new prompt" -- this is an _intended behavior_, whereas the indexing approach creates a potential [unknown unknown](https://en.wikipedia.org/wiki/There_are_unknown_unknowns).

zenodflow:
Closing this PR for now.

PR #18, opened 2024-06-05
feat: add confidence level metric tracking when using CoT and slightly improve benchmark

daiwaid:


libem/match/function.py
```diff
@@ -55,5 +55,22 @@ def func(left, right):
...
-    output = output.split("\n")[-1].lower()
-    return "yes" if "yes" in output else "no"
+    output = [line.lower() for line in output.split("\n")]
+    confidence, answer = None, None
+    
+    # parse answer
+    answer = "yes" if "yes" in output[-1] else "no"
+    
+    # parse confidence score if CoT is True
+    output = list(reversed(output))
```
zenodflow, line 65: I feel we should start using library that gets us structured output, e.g., https://github.com/guidance-ai/guidance or https://github.com/guardrails-ai/guardrails

zenodflow (approved):

PR #19, opened 2024-06-06
refactor: allow model call to take list as prompt

zenodflow:
Changed logic of when to invoke multiple model calls, allow list as input, and return tool outputs.

### DIFF
--- a/libem/core/model.py
+++ b/libem/core/model.py
@@ -26,15 +26,13 @@ def openai(prompt: str | list,
            temperature: float = 0.0,
            seed: int = None,
            max_model_call: int = 3,
+           return_tool_output: bool = False,
            verbose: bool = True,
-           ) -> str:
+           ) -> str or (str, dict):
     if not os.environ.get("OPENAI_API_KEY"):
         raise EnvironmentError(f"OPENAI_API_KEY is not set.")
 
     client = OpenAI()
-
-    num_model_calls = 0
-    num_input_tokens, num_output_tokens = 0, 0
 
     # format the prompt messages
     match prompt:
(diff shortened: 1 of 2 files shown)


libem/core/model.py
```diff
@@ -107,30 +118,32 @@ def openai(prompt: str, tools: list[str] = None,
...
+                    raise libem.ModelTimedoutException(e)
+
+                response_message = response.choices[0].message
+                tool_calls = response_message.tool_calls
+
+                num_model_calls += 1
+                num_input_tokens += response.usage.total_tokens - response.usage.completion_tokens
+                num_output_tokens += response.usage.completion_tokens
+
+            if num_model_calls == max_model_call:
```
daiwaid, line 145: When max_model_call is reached, the while loop does not terminate, so maybe rename the variable to something like model_calls_warn_threshold?

zenodflow: Right, sorry, I fixed the bug in a latest commit by setting the tool_calls to []

zenodflow:
One issue we have is how to get the tool calls' output for later processing .. working on it but open to suggestion @daiwaid

daiwaid:
What do you mean by later processing? Can't we just use trace to log it?

zenodflow:
What if we want to _use_ the output from tool, e.g., invoking libem.match as a tool and use its output?

daiwaid:
The easiest solution would just have the LLM write it out, but that'll require an extra pass to the LLM. Alternatively we can have a list of tools that will output the response and if those tools are invoked then just return the response from it directly rather than passing it back to the LLM.

zenodflow:
I see. How about use a separate return value? E.g., in the latest commit

zenodflow:
@daiwaid I'll merge it for now (it should be backward compatible with the rest of the codebase) and can improve for future.

PR #20, opened 2024-06-07
feat: add challenging dataset

Kathleen0123:


benchmark/run.py
```diff
@@ -8,6 +8,7 @@
     'abt-buy': classic.benchmark_abt_buy,
     'amazon-google': classic.benchmark_amazon_google,
     'beer': classic.benchmark_beer,
+    'challenging': classic.benchmark_challenging,
```
zenodflow, line 11: move this to the bottom of the list

benchmark/classic/__init__.py
```diff
@@ -1,6 +1,7 @@
 from benchmark.classic.abt_buy import benchmark as benchmark_abt_buy
 from benchmark.classic.amazon_google import benchmark as benchmark_amazon_google
 from benchmark.classic.beer import benchmark as benchmark_beer
+from benchmark.classic.challenging import benchmark as benchmark_challenging
```
zenodflow, line 4: move this to the bottom of the import

zenodflow (approved):
Thanks @Kathleen0123 , very nice, we can now start testing on those challenging pairs as a separate dataset. 

Is it possible to add a field in the raw datasets indicating which dataset the data sample is from? Does the original trace contain such information - the file name maybe? 

We want to be able to support "reading challenging cases from X dataset" in addition to read them all. I'll merge this PR first though! Nice work. 

(Don't worry about the comments, I'll address those nits)

PR #26, opened 2024-06-23
feat: add blocking

daiwaid:
- Added blocking implementation using the Levenshtein distance (edit distance)
- Tuned each classic benchmark to support blocking benchmark and found the optimal similarity score for each dataset
- Formatting and cmd argument edits in 1benchmark/run.py1 and 1benchmark/util.py1

### DIFF
--- a/benchmark/classic/dblp_acm.py
+++ b/benchmark/classic/dblp_acm.py
@@ -60,7 +60,7 @@ def run(args):
 
     if args.block:
         libem.calibrate({
-            "libem.block.parameter.similarity": 79
+            "libem.block.parameter.similarity": 81
         })
         
         left = set(json.dumps(d['left']) for d in dataset)
--- a/benchmark/classic/dblp_acm.py
+++ b/benchmark/classic/dblp_acm.py
@@ -1,3 +1,4 @@
+import json
 import random
 
 import libem
@@ -56,5 +57,19 @@ def run(args):
             "libem.match.prompt.experiences": Prompt.Experiences(),
             "libem.match.prompt.output": ""
         })
+
+    if args.block:
+        libem.calibrate({
+            "libem.block.parameter.similarity": 79
+        })
+        
+        left = set(json.dumps(d['left']) for d in dataset)
+        right = set(json.dumps(d['right']) for d in dataset)
+        dataset = {
+            'left': [json.loads(i) for i in left],
+            'right': [json.loads(i) for i in right],
+            'true': [{'left': d['left'], 'right': d['right']} 
+                     for d in dataset if d['label'] == 1]
+        }
 
     util.benchmark(dataset, args)
--- a/benchmark/classic/dblp_acm.py
+++ b/benchmark/classic/dblp_acm.py
@@ -6,6 +6,7 @@ from libem.core.struct import Prompt
 from libem.prepare.datasets import dblp_acm
 
 from benchmark import util
+from benchmark.classic import tuned_similarity
 
 
 def run(args):
@@ -60,7 +61,9 @@ def run(args):
 
     if args.block:
         libem.calibrate({
-            "libem.block.parameter.similarity": 81
+            "libem.block.parameter.similarity": args.similarity
+                                                if 0 <= args.similarity <= 100 
+                                                else tuned_similarity['dblp-acm']
         })
         
         left = set(json.dumps(d['left']) for d in dataset)
--- a/benchmark/classic/walmart_amazon.py
+++ b/benchmark/classic/walmart_amazon.py
@@ -1,3 +1,4 @@
+import json
 import random
 
 import libem
@@ -61,5 +62,19 @@ def run(args):
             "libem.match.prompt.experiences": Prompt.Experiences(),
             "libem.match.prompt.output": ""
         })
+
+    if args.block:
+        libem.calibrate({
+            "libem.block.parameter.similarity": 51
+        })
+        
+        left = set(json.dumps(d['left']) for d in dataset)
+        right = set(json.dumps(d['right']) for d in dataset)
+        dataset = {
+            'left': [json.loads(i) for i in left],
+            'right': [json.loads(i) for i in right],
+            'true': [{'left': d['left'], 'right': d['right']} 
+                     for d in dataset if d['label'] == 1]
+        }
 
     util.benchmark(dataset, args)
--- a/benchmark/classic/walmart_amazon.py
+++ b/benchmark/classic/walmart_amazon.py
@@ -65,7 +65,7 @@ def run(args):
 
     if args.block:
         libem.calibrate({
-            "libem.block.parameter.similarity": 51
+            "libem.block.parameter.similarity": 50
         })
         
         left = set(json.dumps(d['left']) for d in dataset)
--- a/benchmark/classic/walmart_amazon.py
+++ b/benchmark/classic/walmart_amazon.py
@@ -6,6 +6,7 @@ from libem.core.struct import Prompt
 from libem.prepare.datasets import walmart_amazon
 
 from benchmark import util
+from benchmark.classic import tuned_similarity
 
 
 def run(args):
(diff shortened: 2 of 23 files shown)


zenodflow:
@daiwaid Nice. Could you resolve the above conflict over requirements (we should remove sqlite3 - it's built-in)

benchmark/run.py
```diff
@@ -140,6 +143,9 @@ def main():
     parser.add_argument("--confidence", dest='confidence',
                         help="Report confidence score.",
                         action='store_true', default=False)
+    parser.add_argument("--no-match", dest='match',
```
zenodflow, line 146: Readability: Move this argument above with the --block

benchmark/util.py
```diff
@@ -7,6 +7,7 @@
 from datetime import datetime
 
 import libem
+import libem.block
```
zenodflow, line 10: Remove and just use `libem.block()` to reference? Assuming block() is added to libem/interface.py

benchmark/classic/walmart_amazon.py
```diff
@@ -62,4 +63,18 @@ def run(args):
             "libem.match.prompt.output": ""
         })
 
+    if args.block:
+        libem.calibrate({
+            "libem.block.parameter.similarity": 50
```
zenodflow, line 68: Make this configurable with an argument?

benchmark/util.py
```diff
@@ -45,9 +46,139 @@ def benchmark(dataset, args):
...
+        dataset, block_stats, block_results = block(dataset, args)
+        results['block'] = block_results
+    if args.match:
+        match_stats, match_results = match(dataset, args)
+        if args.block:
+            results['match'] = match_results
+        else:
+            results = match_results
+    
+    if args.log:
```
zenodflow, line 60: Is there a reason why we don't always log the benchmark output?

daiwaid:
Done.

requirements.txt
```diff
@@ -15,3 +15,5 @@ fastapi
 pydantic
 uvicorn
 sqlite3
+fuzzywuzzy
+python-Levenshtein
```
zenodflow, line 18: Where is this package used?

benchmark/classic/dblp_acm.py
```diff
@@ -57,4 +58,18 @@ def run(args):
             "libem.match.prompt.output": ""
         })
 
+    if args.block:
+        libem.calibrate({
+            "libem.block.parameter.similarity": 81
```
zenodflow, line 63: Why different settings for different datasets?

zenodflow (commented):
Very nice feature. Left some comments.

- Could you - in a follow-up commit of this PR - come up with a examples/block.py (e.g., takes 5 samples, hard-coded in-file; first block and then match and report useful information in the stages)?
- Could you explain whether / how the benchmark module does block and then match? I.e., does the block stage first dumps the blocks in a batch on disk and then call match over the batch? Or this is per record block and match? 
- Make sure the `make all` still passes all examples.

Then this is good to go!

zenodflow:
> Done.

Instead of doing a merge, could you do a rebase? Let's try merging your entire commit history without squash.

zenodflow:
To add: I think we should (could be a follow-up PR) showcase the benefit of blocking - e.g., add a benchmark and report results in the benchmarks/README.md.

benchmark/util.py
daiwaid: Mostly for when I test the benchmark module and tools, I don't want to clutter my results folder with random runs.

requirements.txt
daiwaid: libem/block/function.py

benchmark/classic/dblp_acm.py
daiwaid: To ensure 100% recall (ensure all positive pairs made through the blocking stage). Think of this as something we have tuned for these datasets, but since the tune module isn't ready I just manually tuned this parameter.

requirements.txt
daiwaid: the python-Levenshtein package is for speeding up the similarity compute (I think this will run the algorithm in C rather than using native python code so it's faster).

daiwaid:
> Could you explain whether / how the benchmark module does block and then match? I.e., does the block stage first dumps the blocks in a batch on disk and then call match over the batch? Or this is per record block and match?

It currently saves the entire potentially matching pairs list in memory, then passes that dataset to match. I think I'll change it to per record (yielding) instead so blocking will only run until num_pairs of potential matches are found.

benchmark/util.py
zenodflow: I see. Makes sense. Default is True, right? By default benchmark should always log.

zenodflow: (so maybe --no-log is what you want)

zenodflow: actually, a cleaner way is set the result dir to /dev/null (which we already have an option for the dir!) if you want to avoid outputing files

requirements.txt
zenodflow: Got it, nice!

benchmark/classic/dblp_acm.py
zenodflow: I am not following: I thought blocking is to prefilter the pairs. Since the pairs we load for these current datasets are already filtered / paired up, why do we need to do blocking here?

Is there a dataset for blocking to test this?

zenodflow:
@daiwaid Thanks for the clarification. I think in reality people run blocking and matching separately in batches so that one can optimize for each. I think you are doing the right thing, just that I think the blocking results should be dumped to disk. Matching should pick up from the list.

If you want to do "block and match" (essentially this is not blocking -- creating pairs -- but filtering), perhaps make a separate API for it (the yielding thing) in `libem.block`. You can put it under the block method (e.g., with a `batch` flag -- when true, the default, does the original blocking; otherwise, return the iterator).

benchmark/util.py
daiwaid: --no-log will change args.log from True to False. Doing it this way avoids having to generate the contents for the output file so it's a bit faster vs just outputting to a null file.

benchmark/classic/dblp_acm.py
daiwaid: The original dataset of pairs is split up into a left dataset and a right dataset, then passed through blocking. This similarity number is to ensure that all the original positive pairs (label = 1) are still there after the blocking phase.

zenodflow: I see. Is it really necessary? Are we not biasing the results? Also, is the splitting happen at runtime? Or is there a separate source dataset your prepared for libem-sample-data?

Meaning, if we want to evaluate blocking, I think we should evaluate it correctly and fairly. However, I agree that manually figuring out the similarity threshold and put it as place holder is good. Perhaps remember to print the threshold out at the result -- i.e., make the similarity threshold a dict in the classic/__init__.py or somewhere. Not so much as a hard-coded value buried in.

benchmark/util.py
zenodflow: actually, a cleaner way is set the result dir to /dev/null (which we already have an option for the dir!) if you want to avoid outputing files

daiwaid: It does output log by default, args.log is default to True and --no-log flag will set it to False.

benchmark/classic/dblp_acm.py
daiwaid: The splitting is done during runtime. I believe this is the correct way to evaluate blocking performance since the goal of  blocking is to preserve as many true match pairs as possible while limiting the amount of false positive pairs. Maybe we don't need it to have a 100% recall accuracy but any penalty we take during the blocking phase will cap the maximum performance of matching (no way to get 100% recall during matching if some pairs never make it to that stage).

zenodflow:
LGTM. Would be nice to have 1. a block example; 2. a rebase to get rid of "[Merge branch 'main' into feat-blocking](https://github.com/abcsys/libem/pull/26/commits/4b2152d046ca46449ef1dc2c4451c212fa4ac51a)"

daiwaid:
@zenodflow Take a look at the new changes, I also reworked the commits.

> just that I think the blocking results should be dumped to disk. Matching should pick up from the list.

I think this decision should be made by the client-side code, and for the benchmarking module, this is unnecessary.

daiwaid:
Final changes have been applied and bumped version number to 0.16.

zenodflow (approved):

zenodflow:
Thanks! @daiwaid 

> bumped version number to 0.16.

In future, it would be nice to have the version bump in its own commit.

PR #28, opened 2024-06-26
feat: page routing, collape result popup, save session token to cookies

daiwaid:
- add page routing (able to back/forward)
- add ability to collapse match result popup to the bottom of the screen
- saves session token (uuid) in cookies for up to 40 days


daiwaid:
PR changes are now live at http://arena.libem.org

zenodflow:
Very nice. The design is neat!

arena/src/Common.js
```diff
@@ -0,0 +1,33 @@
+import { v4 as uuidv4 } from 'uuid'
+
+const baseURL = "http://127.0.0.1:8000"
```
zenodflow, line 3: We should support https / certificate in future; some browser does not open http tab nowadays.

zenodflow (approved):
Lgtm

PR #30, opened 2024-07-01
Local model support

Char15Xu:
Added local model support for apple silicon devices

### DIFF
--- a/examples/local.py
+++ b/examples/local.py
@@ -1,0 +1,33 @@
+import libem
+from libem.match.prompt import rules
+
+def positive():
+    e1 = "Dyson Hot+Cool AM09 Jet Focus heater and fan, White/Silver"
+    e2 = "Dyson AM09 Hot + Cool Jet Focus Fan Heater - W/S"
+
+    is_match = libem.match(e1, e2)
+
+    print("Entity 1:", e1)
+    print("Entity 2:", e2)
+    print("Match:", is_match['answer'])
+
+def negative():
+    e1 = "Dyson Hot+Cool AM09 Jet Focus heater and fan, White/Silver"
+    e2 = "Dyson AM09 Hot + Cool Jet Focus Fan Heater - Black japan"
+
+    rules.add("Color differentiates entities.")
+    is_match = libem.match(e1, e2)
+
+    print("Entity 1:", e1)
+    print("Entity 2:", e2)
+    print("Match:", is_match['answer'])
+
+def main():
+    libem.calibrate({
+        "libem.match.parameter.model": "llama3",
+    }, verbose=True)
+    positive()
+    negative()
+
+if __name__ == '__main__':
+    main()
--- a/libem/core/model.py
+++ b/libem/core/model.py
@@ -8,7 +8,14 @@ import libem
 
 
 def call(*args, **kwargs) -> dict:
-    return openai(*args, **kwargs)
+    if 'model' in kwargs:
+        model_type = kwargs['model']
+        if model_type in ["gpt-4o", "gpt-4-turbo", "gpt-3.5-turbo"]:
+            return openai(*args, **kwargs)
+        elif model_type == "llama3":
+            return local(*args, **kwargs)
+        else:
+            raise ValueError(f"Invalid model: {model_type}")
 
 
 """ OpenAI """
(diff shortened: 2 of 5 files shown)


libem/core/model.py
```diff
@@ -177,3 +184,52 @@ def openai(prompt: str | list | dict,
...
+from mlx_lm import load, generate
+def local(prompt: str | list | dict,
+           tools: list[str] = None,
+           context: list = None,
+           model: str = "null",
+           temperature: float = 0.0,
+           seed: int = None,
+           max_model_call: int = 3,
+           ) -> dict:
+    import time
```
zenodflow, line 198: Make it to top import

libem/core/model.py
```diff
@@ -177,3 +184,52 @@ def openai(prompt: str | list | dict,
...
+        case str():
+            messages = [{"role": "user", "content": prompt}]
+        case _:
+            raise ValueError(f"Invalid prompt type: {type(prompt)}")
+
+    context = context or []
+    input_text = '\n'.join(f"{prompt['role']}: {prompt['content']}" for prompt in messages)
+    messages = context + [input_text]
+
+    if not tools:
```
zenodflow, line 226: If the current local model does not support tool use, but tools are provided, let's raise a new ToolUseUnsupported error

libem/parameter.py
```diff
@@ -1,8 +1,8 @@
 from libem.core.struct import Parameter
 
 model = Parameter(
-    default="gpt-4o",
-    options=["gpt-4-turbo", "gpt-3.5-turbo"]
+    default="llama3",
```
zenodflow, line 4: Make default gpt-4o and add llama3 in options

zenodflow (changes requested):
@Char15Xu nice work! Could we add an examples/local.py to showcase the libem.match with local model?

Use the libem.calibrate tool to set the model to llama3

Char15Xu:
Added the local.py file to showcase with local model, fixed all three issues in the comments.

libem/core/model.py
```diff
@@ -177,3 +184,54 @@ def openai(prompt: str | list | dict,
...
+           context: list = None,
+           model: str = "null",
+           temperature: float = 0.0,
+           seed: int = None,
+           max_model_call: int = 3,
+           ) -> dict:
+
+    # Load the model using MLX for apple silicon device
+    if model == "llama3":
+        model_path = "mlx-community/Meta-Llama-3-8B-Instruct-4bit"
```
zenodflow, line 201: Since there is this dependency, I think we should add a Makefile entry to help users to prepare this?

libem/core/model.py
```diff
@@ -177,3 +184,54 @@ def openai(prompt: str | list | dict,
...
+    context = context or []
+    input_text = '\n'.join(f"{prompt['role']}: {prompt['content']}" for prompt in messages)
+    messages = context + [input_text]
+
+    if not tools:
+        try:
+            response = generate(model_local, tokenizer, prompt=messages[0], max_tokens=10, temp=temperature)
+        except APITimeoutError as e:  # catch timeout error
+            raise libem.ModelTimedoutException(e)
+    else:
```
zenodflow, line 231: Make this:

`if tools: .. else: ..`

examples/local.py
```diff
@@ -0,0 +1,33 @@
+import libem
+from libem.match.prompt import rules
+
+def positive():
```
zenodflow, line 4: This is cool! Could you add a `make local` to run this example?

zenodflow (commented):
Lgtm. Only minor stylish comments and we're good to go! Nice work @Char15Xu.

libem/core/model.py
Char15Xu: mlx-lm is integrated with huggingface, so properly installing mlx and mlx-lm packages allows the load() function to download the model automatically.

libem/core/model.py
Char15Xu: do you mean `if not tools: ... else: ...`

zenodflow: If tools: raise ...
else: ...

zenodflow (approved):

zenodflow:
@Char15Xu, could you push this to a `feat-local` branch?

zenodflow:
The `core/model.py` needs to handle different platforms. One possibility is to put the mlx dependency in a separate `lite` directory or a `libem-lite` repo.

Char15Xu:
should i switch back to llama.cpp?

zenodflow:
For apple silicon we can use mlx and then the other platforms for llama.cpp

PR #31, opened 2024-07-03
Revert "Local model support"

zenodflow:
Reverts abcsys/libem#30

Unable to resolve Mlx dependency on none apple silicon, we should reconsider how to support local mode without interfering compatibility.


Char15Xu:
should i swtich back to llama.cpp? It works on all platforms.

PR #32, opened 2024-07-04
feat: async model call and match (old)

daiwaid:


libem/match/function.py
```diff
@@ -1,17 +1,20 @@
...
 from libem.core import struct
 from libem.core import model
+from libem.core.util import throttled_async_run_all
 
 schema = {
     "type": "function",
     "function": {
         "name": "match",
-        "description": "Perform entity matching given two entity descriptions.",
+        "description": "Perform entity matching given a pair or two lists of entity descriptions.",
```
daiwaid, line 17: Unintended change, will remove later.

zenodflow:
Thanks @daiwaid. The async implementation looks good to me overall. Two proposals:
* We should preserve the synchronous APIs. The sync and async should be in separate code path.
* This PR should probably wait until we figure out the batch match (multiple questions - answers in an input prompt) performance.

libem/core/model.py
```diff
@@ -1,14 +1,16 @@
...
+import httpx
 import importlib
 
-from openai import OpenAI, APITimeoutError
+from openai import AsyncOpenAI, APITimeoutError
 
 import libem
 
 
-def call(*args, **kwargs) -> dict:
```
zenodflow, line 10: We should keep the original sync call. Perhaps make it an async wrapper with asyncio runner. The async call should be explicit, e.g., async_call

libem/core/model.py
```diff
@@ -18,9 +20,17 @@ def call(*args, **kwargs) -> dict:
...
+    http_client=httpx.AsyncClient(
+        limits=httpx.Limits(
+            max_connections=1000,
+            max_keepalive_connections=100
+        )
+    )
+)
 
 # LLM call with multiple rounds of tool use
-def openai(prompt: str | list | dict,
```
zenodflow, line 23: Likewise - make the sync call a wrapper.

libem/core/model.py
```diff
@@ -52,14 +60,14 @@ def openai(prompt: str | list | dict,
 
     # trace variables
     num_model_calls = 0
-    num_input_tokens, num_output_tokens = 0, 0
-    tool_outputs = {}
+    input_tokens, output_tokens = 0, 0
+    tools_used = []
```
zenodflow, line 64: what about tool outputs?

libem/core/model.py
```diff
@@ -31,8 +41,6 @@ def openai(prompt: str | list | dict,
     if not os.environ.get("OPENAI_API_KEY"):
         raise EnvironmentError(f"OPENAI_API_KEY is not set.")
 
-    client = OpenAI()
```
zenodflow, line 34: Is there a reason why making the client a global variable

libem/core/model.py
daiwaid: It's renamed to tools_used, and all original tool_outputs are now appended as part of tools_used.

libem/core/model.py
zenodflow: I think: make the openai() a sync wrapper, use factory method for the client (for async only)

libem/core/model.py
zenodflow: What is inside tools_used? Would it be confusing to mix the two up?

zenodflow: Also, what is the reason for tracking tools_used?

libem/core/model.py
daiwaid: Multiple clients lead to memory leaks or timeouts: https://github.com/openai/openai-python/issues/874

zenodflow:
For naming: let's name all async methods explicitly with an `async_` prefix, e.g., `async_match`. Its sync wrapper should name as-is.

We should get an idea of performance (tput) improvement with `async` first, also the overhead of asyncio loop init in the sync call.

libem/core/model.py
daiwaid: `tool_outputs` was only used for trace, but the redesign changes the `tool_outputs` as part of the returned dict instead, so now all `tool_outputs` are kept in `tools_used` until the function is returned.

libem/core/model.py
zenodflow: yeah, which is why perhaps using a factory method?

zenodflow:
Note that vs. #34 (query batching, async support is needed because query batching does not work well when there is tool use (e.g., browsing) or more complex match outputs.

zenodflow:
With #34 in, we should decide the necessity for async match (async model call is still well motivated, though).

daiwaid:
Continued in #37.

PR #33, opened 2024-07-04
Fix: trace changes in libem.match and libem.core.model

daiwaid:


zenodflow:
@daiwaid could we limit the changes to just `feat: return stats in model call` and nothing else? E.g., not renaming `num_input_tokens` etc. (unless there is a strong reason why), not changing `tool_outputs`.

Let's make these core API changes one at a time.

zenodflow:
The renaming of `entity_1` to `left` etc. is fine -- it's more of an external API.

libem/match/function.py
```diff
@@ -63,19 +63,20 @@ def func(left, right) -> dict:
...
                 f"{pformat(_prompt, sort_dicts=False)}\n"
                 f"[match] model output:\n"
-                f"{model_output}")
+                f"{model_output['output']}")
 
-    output = parse_output(model_output)
+    output = parse_output(model_output["output"])
 
-    libem.trace.add({"match": {"left": left, "right": right,
+    libem.trace.add({"match": {"id": id,
```
zenodflow, line 75: Why do we need this ID?

libem/core/model.py
```diff
@@ -126,13 +125,11 @@ def openai(prompt: str | list | dict,
                     }
                 )
 
-                libem.trace.add({
```
zenodflow, line 129: Why removing this?

libem/core/model.py
```diff
@@ -126,13 +125,11 @@ def openai(prompt: str | list | dict,
...
-                    'tool': {
-                        "id": tool_call.id,
-                        'name': function_name,
-                        "arguments": function_args,
-                        "response": function_response,
-                    }
+                tools_used.append({
```
zenodflow, line 128: This should be called tool_outputs

libem/core/model.py
```diff
@@ -164,16 +161,24 @@ def openai(prompt: str | list | dict,
 
     messages.append(response_message)
 
+    stats = {
+        "model_calls": num_model_calls,
```
zenodflow, line 165: this should be called num_model_calls

(and num_input_tokens and num_output_tokens; also the variables should be named likewise!)

libem/core/model.py
daiwaid: Each tool should log its own trace, so this is unnecessary. All tool calls will be logged at the end as a single trace.

benchmark/util.py
```diff
@@ -222,30 +222,26 @@ def benchmark_match(dataset, args):
             # get unparsed model output and telemetry
             latency = time.time() - start_time
 
-            model_output = [i['match']['model_output'] for i in t.get() if 'match' in i]
+            output = [i['match'] for i in t.get() if 'match' in i][0]
```
zenodflow, line 225: What is [0]? Reading off this code, I have no idea what [0] is and there is no comment also

benchmark/util.py
```diff
@@ -222,30 +222,26 @@ def benchmark_match(dataset, args):
...
             results.append({
-                'entity_1': e1,
-                'entity_2': e2,
+                'left': output['left'],
+                'right': output['right'],
                 'label': label,
-                'pred': is_match["answer"],
-                'confidence': is_match["confidence"],
-                'explanation': is_match["explanation"],
+                'pred': output['result']['answer'],
```
zenodflow, line 239: We really shouldn't make the match output (answer, confidence, explanation) nested. It should be as simple as possible.

libem/core/model.py
zenodflow: no. the log you are talking about are within individual tool, not the one which was here (e.g., a within a tool it won't know anything about the id)

benchmark/util.py
daiwaid: There will only be 1 "match" trace per match() call, so 0 is just to get the only item out of the loop comprehension.

libem/core/model.py
```diff
@@ -164,16 +161,24 @@ def openai(prompt: str | list | dict,
...
             "num_output_tokens": num_output_tokens,
+            "tool_outputs": tool_outputs,
         }
     })
 
     return {
         "output": response_message.content,
         "messages": messages,
         "tool_outputs": tool_outputs,
+        "usage": usage
```
zenodflow, line 183: just `**usage`, let's not nest them

libem/core/model.py
```diff
@@ -164,16 +161,24 @@ def openai(prompt: str | list | dict,
...
+    usage = {
+        "num_model_calls": num_model_calls,
+        "num_input_tokens": num_input_tokens,
+        "num_output_tokens": num_output_tokens,
+    }
+    
     libem.trace.add({
         "model": {
             "num_model_calls": num_model_calls,
             "num_input_tokens": num_input_tokens,
```
zenodflow, line 173: `**usage`

libem/match/function.py
```diff
@@ -63,19 +63,19 @@ def func(left, right) -> dict:
         model=parameter.model(),
         temperature=parameter.temperature(),
         seed=libem.LIBEM_SEED,
-    )["output"]
```
zenodflow, line 66: rename variable `model_output` to `response`

libem/core/model.py
daiwaid: This id isn't useful to us, it's only for cross-referencing the output from the model call. These tool results are still being logged, just at the same time as the model trace at the end of this file (which is better imo since the model output will also be there).

benchmark/util.py
zenodflow: I knew that. My point was, [0] is hard to understand.

libem/core/model.py
zenodflow: The point is not just about the id but the `tool` span. You removed it entirely.

daiwaid:
Addressed all comments in last commit.

zenodflow:
Nice - let me do a final pass.

zenodflow:
Merge for now - changes in a follow-up PR.

PR #34, opened 2024-07-05
feat: prompt-level batching for match

daiwaid:
The normal (single pair) matching code in libem/match.py and benchmark/util.py are left as is in anticipation of rework in PR #32. The TODOs in benchmark/util.py require changes from PR #33.

### DIFF
--- a/examples/optimize/batch.py
+++ b/examples/optimize/batch.py
@@ -1,8 +1,0 @@
-import subprocess
-from libem.match.parameter import batch_size
-
-print("Without batching:")
-subprocess.run(['python', '-m', 'benchmark.run', '-p', '10', '--no-shuffle', '--no-log', '-q'])
-print()
-print(f"With batching (batch size of {batch_size()}):")
-subprocess.run(['python', '-m', 'benchmark.run', '-p', '10', '--batch', '--no-shuffle', '--no-log', '-q'])
--- a/examples/optimize/batch.py
+++ b/examples/optimize/batch.py
@@ -1,0 +1,8 @@
+import subprocess
+from libem.match.parameter import batch_size
+
+print("Without batching:")
+subprocess.run(['python', '-m', 'benchmark.run', '-p', '10', '--no-shuffle', '--no-log', '-q'])
+print()
+print(f"With batching (batch size of {batch_size()}):")
+subprocess.run(['python', '-m', 'benchmark.run', '-p', '10', '--batch', '--no-shuffle', '--no-log', '-q'])
--- a/libem/match/function.py
+++ b/libem/match/function.py
@@ -119,8 +119,14 @@ def batch(left: list, right: list) -> list[dict]:
 
         batch_start += batch_size
 
+    num_batches = len(left_batches)
+
     output = []
-    for l, r in zip(left_batches, right_batches):
+    for i, (l, r) in enumerate(zip(left_batches, right_batches)):
+        if not parameter.quiet():
+            libem.info(f"[match] processing batch "
+                       f"{i + 1} / {num_batches} "
+                       f"of size {len(l)}.")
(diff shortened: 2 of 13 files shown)


libem/match/function.py
```diff
@@ -30,7 +32,34 @@
 }
 
 
-def func(left, right) -> dict:
+def func(left: Any | list[Any], 
+         right: Any | list[Any]) -> Any | list[Any]:
+    if parameter.batch():
+        assert len(left) == len(right)
+        left_batches, right_batches, batch_ids = [], [], []
```
zenodflow, line 39: What are the assumptions regarding the left_batches and the right_batches? E.g., do they need to be paired up and if so, perhaps add some checks here.

zenodflow:
Nice! This is a key feature. @daiwaid Could you add an example in `examples/optimize/batch.py`? I will also use it to run some testing on your branch to help the review. Thanks!

In the example, perhaps do just 10 pairs to avoid burning tokens. Also, print out the stats so that we now about the latency etc.

daiwaid:
Added example, and deal with case when model will only output a single answer for an entire batch.

examples/optimize/batch.py
```diff
@@ -0,0 +1,7 @@
+import subprocess
+
+print("Without batching:")
+subprocess.run(['python', '-m', 'benchmark.run', '-p', '10', '--no-shuffle', '--no-log', '-q'])
+
+print("With batching:")
+subprocess.run(['python', '-m', 'benchmark.run', '-p', '10', '--batch', '--no-shuffle', '--no-log', '-q'])
```
zenodflow, line 7: Specify batch_size

libem/match/function.py
zenodflow: I think these data preprocessing should be placed within the batch_match not here. The batch_match can have its own "inner loop" to process micro batches.

libem/match/function.py
```diff
@@ -79,6 +108,73 @@ def func(left, right) -> dict:
...
     return output
 
+def batch_match(left: list[Any], right: list[Any], ids: list[int] = []) -> list[dict]:
+    start = time.time()
+    output = []
+    
+    system_prompt = Prompt.join(
+        prompt.role(),
+        prompt.rules(),
+        prompt.experiences(),
```
zenodflow, line 118: I don't think the batch mode supports all these features? If so, let's remove all these other prompts, for now.

libem/match/parameter.py
```diff
@@ -16,3 +16,11 @@
 confidence = Parameter(
     default=False
 )
+
+# model query batching
```
zenodflow, line 20: `match query batching`

libem/match/function.py
daiwaid: This is to be consistent with the async match changes. I don't think an additional function is necessary here since the outermost func() is only acting as a coordinator. And the code is cleaner this way (I started with batch_match() handling batching).

zenodflow:
Please check #35

libem/match/function.py
zenodflow: What I meant was: keep this func() as a thin wrapper and put all batch processing in batch_match

daiwaid:
Merged new changes from #33, #35 and cleaned up commit history to align with main branch.

zenodflow:
This should be now good to go.

Keeping a note on the performance improvements:

```
Benchmark: Matching done in 107.63s.
Benchmark: Precision	 66.57
Benchmark: Recall	 93.59
Benchmark: F1 score	 77.8
Benchmark: Cost 	 $0.53134
```

This is on the `amazon-google` dataset; we see roughly 7x faster completion, 3% higher F1, with half of the cost.

zenodflow (approved):
Lgtm.

PR #37, opened 2024-07-08
feat: async model call and match

daiwaid:


daiwaid:
| Batch Size | F1    | Runtime Latency | Per Pair Latency | Cost |
|:----------:|:-----:|:---------------:|:----------------:|:----:|
| 1          | 94.37 | 14.44           | 2.44             | 1.20 |
| 4          | 93.78 | 3.65            | 0.41             | 1.06 |
| 16         | 94.06 | 11.60           | 0.15             | 1.01 |
| 64         | 94.86 | 11.98           | 0.12             | 0.99 |
| 128        | 94.54 | 22.32           | 0.11             | 0.99 |
| 256        | 93.43 | 32.93           | 0.10             | 0.99 |
| 512        | 70.87 | 74.44           | 0.08             | 0.95 |

zenodflow:
> Batch Size	F1	Runtime Latency	Per Pair Latency	Cost
> 1	94.37	14.44	2.44	1.20
> 4	93.78	3.65	0.41	1.06
> 16	94.06	11.60	0.15	1.01
> 64	94.86	11.98	0.12	0.99
> 128	94.54	22.32	0.11	0.99
> 256	93.43	32.93	0.10	0.99
> 512	70.87	74.44	0.08	0.95

Interestingly, the F1 has a sharp drop at 512.

libem/match/function.py
```diff
@@ -98,41 +109,52 @@ def once(left: str, right: str) -> dict:
...
+    return await throttled_async_run_all(
+        once(l, r)
+        for l, r in zip(left, right)
+    )
+
+
+async def batch(left: list, right: list) -> list[dict]:
     assert len(left) == len(right)
+    
+    if len(left) <= parameter.batch_size():
```
daiwaid, line 122: This is for a bit of performance speed up (if only 1 batch, don't need to run the _create_batches() code and don't need a progress bar) and to allow for the "detailed mode" in benchmark where entities and predictions are printed as they are matched.

benchmark/util.py
```diff
@@ -167,7 +172,7 @@ def run_block(dataset, args):
         'precision': round(precision, 2),
         'recall': round(recall, 2),
         'f1': f1,
-        'latency': total_time,
+        'runtime_latency': total_time,
```
zenodflow, line 175: why not just `latency`?

benchmark/util.py
```diff
@@ -197,81 +202,123 @@ def run_block(dataset, args):
...
+        labels.append(data['label'])
 
-                if not args.quiet:
+    with libem.trace as t:
+        libem.calibrate({
+                "libem.match.parameter.batch_size": args.batch_size,
+            })
+        
+        # no async
+        if args.info or args.debug:
```
zenodflow, line 236: this is very confusing

zenodflow: why no async relate to args.info and args.debug

benchmark/util.py
```diff
@@ -197,81 +202,123 @@ def run_block(dataset, args):
...
+                                f"due to model call timeout..")
+            else:
+                # generate left and right batches
+                num_pairs = len(left_set)
+                batch_start, batch_index = 0, 0
+                batches = []
+
+                while batch_start < num_pairs:
+                    batch_end = min(batch_start + args.batch_size, num_pairs)
+                    batches.append((batch_index,
```
zenodflow, line 279: why the batch preparation code is here?

benchmark/util.py
```diff
@@ -197,81 +202,123 @@ def run_block(dataset, args):
...
+        if args.info or args.debug:
+            if args.batch_size == 1:
+                # iterate over dataset
+                for i, left, right, label in zip(range(len(labels)), 
+                                                  left_set, right_set, labels):
                     print(f"Pair #{i + 1}\n")
                     print(f"Entity 1: {left}\n")
                     print(f"Entity 2: {right}")
 
-                num_retries = 0
```
zenodflow, line 226: revert to old

benchmark/util.py
zenodflow: revert to the original one

benchmark/util.py
```diff
@@ -197,81 +202,123 @@ def run_block(dataset, args):
...
+                    batch_end = min(batch_start + args.batch_size, num_pairs)
+                    batches.append((batch_index,
+                                   left_set[batch_start:batch_end],
+                                   right_set[batch_start:batch_end],
+                                   labels[batch_start:batch_end]))
+                    batch_start += args.batch_size
+                    batch_index += 1
+                
+                # iterate over batches
+                for i, left, right, label in batches:
```
zenodflow, line 287: this is client code, it should not have the notion of batches

benchmark/util.py
```diff
@@ -197,81 +202,123 @@ def run_block(dataset, args):
...
-                    except libem.ModelTimedoutException:
-                        num_retries += 1
-                        print(f"Retrying {num_retries} time(s) "
-                              f"due to model call timeout..")
+                        truth.append(label[i])
+        
+        # with async
         else:
-            # batch matching
-            libem.calibrate({
```
zenodflow, line 255: this is no calibration code for async?

benchmark/util.py
```diff
@@ -331,7 +378,8 @@ def run_match(dataset, args):
         'precision': round(metrics['precision'] * 100, 2),
         'recall': round(metrics['recall'] * 100, 2),
         'f1': round(metrics['f1'] * 100, 2),
-        'latency': round(time.time() - start_time, 2),
+        'runtime_latency': round(time.time() - start_time, 2),
```
zenodflow, line 381: this should be just `latency`

examples/optimize/prompt_batch.py
```diff
@@ -13,6 +13,9 @@ def main():
         samples.append(next(dataset))
 
     print("Without batch:")
+    libem.calibrate({
+        "libem.match.parameter.batch_size": 1,
```
zenodflow, line 17: good change - improves readability

libem/constant.py
```diff
@@ -24,4 +24,6 @@
 LIBEM_3RD_PARTY_LOG_LEVEL = \
     os.environ.get("LIBEM_3RD_PARTY_LOG_LEVEL", logging.WARNING)
 
+LIBEM_MAX_ASYNC_TASKS = 500
```
zenodflow, line 27: maybe put this to `match`. putting it here would mean globally this is the limit across all tools

libem/core/model.py
```diff
@@ -18,6 +23,14 @@ def call(*args, **kwargs) -> dict:
...
+    http_client=httpx.AsyncClient(
+        limits=httpx.Limits(
+            max_connections=1000,
+            max_keepalive_connections=100
+        )
+    )
+)
 
 # LLM call with multiple rounds of tool use
 def openai(prompt: str | list | dict,
```
zenodflow, line 36: *args **kwargs

the signature is in the later async call

libem/core/model.py
```diff
@@ -28,11 +41,22 @@ def openai(prompt: str | list | dict,
...
+           context: list = None,
+           model: str = "gpt-4o",
+           temperature: float = 0.0,
+           seed: int = None,
+           max_model_call: int = 3,
+           ) -> dict:
     if not os.environ.get("OPENAI_API_KEY"):
         raise EnvironmentError(f"OPENAI_API_KEY is not set.")
 
-    client = OpenAI()
```
zenodflow, line 34: not sure if the client is lazily setup, if not, pls do a factory method

libem/core/util.py
```diff
@@ -30,3 +35,20 @@ def get_func(full_path):
         raise Exception(f"Module '{module_path}' does not have a function named '{function_name}': {e}")
     except Exception as e:
         raise Exception(f"An error occurred: {e}")
+
+async def throttled_async_run_all(tasks: Iterable,
```
zenodflow, line 39: such a long method name. this is just `async_run`. your signature already indicates it will throttle. 

rename `limit` to `max_concurrent`

libem/interface.py
```diff
@@ -38,11 +37,13 @@ def chat(message, context=None) -> dict:
...
 
 
-def block(left: list, right: list) -> Iterator[dict]:
+def block(left: list[str | dict], 
+          right: list[str | dict]) -> Iterator[dict]:
     yield from block_func(left, right)
 
 
-def match(left: str | list, right: str | list) -> dict | list[dict]:
+def match(left: Any | list[Any],
```
zenodflow, line 45: keep the specific type, not Any type.

libem/interface.py
```diff
@@ -98,7 +99,11 @@ def config():
     )
 
 
-def debug_on():
+def info_mode():
```
zenodflow, line 102: what is the need for `info_mode`?

libem/interface.py
```diff
@@ -98,7 +99,11 @@ def config():
...
+    libem.LIBEM_LOG_LEVEL = logging.INFO
+    
+
+def debug_mode():
```
zenodflow, line 106: `debug_mode()` sounds like a getter but here is a setter, consider `debug_on`

libem/match/function.py
```diff
@@ -32,16 +35,24 @@
 }
 
 
-def func(left: str | list, right: str | list) -> dict | list[dict]:
+def func(left: Any | list[Any],
```
zenodflow, line 38: use specific type

zenodflow: if your point is that any types can be parsed in, then you must explicitly do type casting later.

otherwise, let the _client code_ do the stringification

libem/match/function.py
```diff
@@ -32,16 +35,24 @@
...
+                     right: Any | list[Any]) -> dict | list[dict]:
     assert type(left) == type(right)
 
     if parameter.batch_size() > 1:
-        return batch(left, right)
+        return await batch(left, right)
     else:
-        return once(left, right)
+        if type(left) is list:
+            return await individual(left, right)
```
zenodflow, line 50: confusing name `individual`.

note that naming like `once` is conventional

libem/match/function.py
zenodflow: where is the progress bar? a progress bar cannot exist inside the core match code

zenodflow: how much of a performance speed up? if not significant (e.g., more than 5%), let's keep it simple and dumb

libem/match/function.py
```diff
@@ -98,41 +109,52 @@ def once(left: str, right: str) -> dict:
...
-    output = []
-    for i, (l, r) in enumerate(zip(left_batches, right_batches)):
-        if not parameter.quiet():
-            libem.info(f"[match] processing batch "
-                       f"{i + 1} / {num_batches} "
-                       f"of size {len(l)}.")
-        output.extend(_proc_batch(l, r))
-
-    return output
+    return batches
```
zenodflow, line 154: the original logic is clearer

zenodflow (changes requested):
Nicely done @daiwaid -- this is impressive speed up! Just need some polishing over the implementation.

libem/match/parameter.py
```diff
@@ -21,7 +21,3 @@
 batch_size = query_per_prompt = Parameter(
     default=1
 )
-
-quiet = Parameter(
```
zenodflow, line 25: what was this originally for?

zenodflow: ah, I seem to recall.

the log level info / debug / warning etc. should not be directly translated to the client code as `info mode` etc.

libem/match/function.py
daiwaid: The progress bar is in the throttled_async_run_all() function in libem.core.util

benchmark/util.py
daiwaid: It's to differentiate this latency from the `avg_per_pair_latency` which is not `latency / len(pairs)` but the avg of all individual pair latencies. Basically `runtime_latency` takes into account the async optimization but `avg_pair_latency` doesn't.

libem/constant.py
daiwaid: This constant is used in `libem.core.util`. It is meant to be a global limit. There is currently no match specific limit, but if further testing suggests that a lower limit is better then we can add one later.

libem/match/function.py
daiwaid: This is meant to be the opposite of `batch()`, where all pairs are matched individually.

benchmark/util.py
```diff
@@ -197,36 +200,49 @@ def run_block(dataset, args):
...
+    start_index = max(args.start_index or 0, 0)
 
     results = {}
     truth, predictions, = [], []
+    total_latency = 0
 
-    print(f"Benchmark: Matching {args.num_pairs if args.num_pairs > 0 else 'all'} "
-          f"{'pair' if args.num_pairs == 1 else 'pairs'} "
+    if args.num_pairs > 0:
+        num_pairs = min(args.num_pairs, len(dataset) - start_index)
```
zenodflow, line 210: this is very confusing

libem/core/model.py
```diff
@@ -1,37 +1,58 @@
...
+    # retrieve api key and set up openai client
+    global openai_client
+    
     if not os.environ.get("OPENAI_API_KEY"):
-        raise EnvironmentError(f"OPENAI_API_KEY is not set.")
-
-    client = OpenAI()
+        os.environ.setdefault(
+            "OPENAI_API_KEY",
+            libem.LIBEM_CONFIG.get("OPENAI_API_KEY", "")
```
zenodflow, line 43: What if LIBEM_CONFIG has no OPENAI_API_KEY set? 

This logic should not be handled here.

zenodflow:
Asynchronous tool use will be introduced in future PRs. For this PR, async calls are used only internally within each tool

zenodflow:
https://github.com/openai/openai-python/issues/1059

Note that reusing AsyncOpenAI client can lead to misc issues. The one observed is that when the concurrency level differs > ~3, we see `openai.APIConnectionError: Connection error.` for consecutive `libem.match` calls.

zenodflow:
Module-level client also discouraged: https://github.com/openai/openai-python/issues/820

zenodflow (approved):

PR #39, opened 2024-07-10
feat: add support for local model

Char15Xu:


Makefile
```diff
@@ -13,13 +13,15 @@ uninstall:
 run: match
 match:
 	python examples/match.py
+match-local:
```
zenodflow, line 16: just `local`; don't make it part of `all` for now; add local to .PHONY

libem/optimize/cost/openai.py
```diff
@@ -25,6 +25,8 @@ def get_output_cost(model, num_tokens):
 
 
 def get_cost(model, num_input_tokens, num_output_tokens):
+    if model == "llama3":
```
zenodflow, line 28: This module is for openai model, llama3 should not appear here. It is the client code that deal with get_cost need to use the right method. To make this easy for client code, we can add a get_cost wrapper for the cost module.

zenodflow:
Follow-up for #30 #31.

Makefile
```diff
@@ -19,7 +19,7 @@ chat:
 	python examples/tool/chat.py
 block:
 	python examples/block.py
-all: match browse chat
+all: match match-local browse chat
```
zenodflow, line 22: remove match-local

zenodflow:
Nice work. I will do a pass and merge.

PR #52, opened 2024-07-25
feat: always pass train set as part of benchmark

daiwaid:


benchmark/classic/abt_buy.py
```diff
@@ -35,12 +35,10 @@ def run(args):
     kwargs.update(args.kwargs or {})
 
     # get dataset with kwargs
-    if args.train:
-        dataset = list(abt_buy.read_train(**kwargs))
-    else:
-        dataset = list(abt_buy.read_test(**kwargs))
+    train_set = list(abt_buy.read_train(**kwargs))
```
zenodflow, line 38: @daiwaid Perhaps not materialize the train_set immediately here until later `if args.train` is True? We are not processing train_set data anyway here. In fact, only materialize the test_set under the `args.shuffle`?

daiwaid: `args.train` is no longer an option because it is incompatible with few-shot, and will potentially make the return statement ugly. I think this is fine because the purpose of the benchmark tool is to test the performance of the model and the training set is used for training not testing. We need to materialize the dataset immediately to get the total number of items in the dataset for print statements in `benchmark.util`.

zenodflow: Speaking of "return statement ugly": the parameter should be train, test not dict.

zenodflow: My point was don't materialize the datasets until its first use, not immediately.

daiwaid: This only defeats the purpose of returning an iterator for the benchmark module, any other code that uses the datasets can choose to keep it as an iterator. I think the current way is better because 1. the return value will always be a list rather than a list when shuffled and an iterator when not (helps with readability) 2. the dataset will be materialized immediately at the start of `benchmark.util`, and since only `benchmark.util` uses these functions I think it makes no difference where to materialize them.

zenodflow: I hear you. But shouldn't the returned datasets, for train and for test, are Iteratable? List or not feels like a runtime detail?

daiwaid: The big difference is that we can immediately get the length of a list but have to iterate through to get the length for a generic iterable, and one of the first things the benchmark module does is get the length of the dataset.

zenodflow: Let me try to understand: you are saying that because the first thing the benchmark module (the caller / client code) does is to print the length, so the dataset reading code (the callee) should materialize the list? 

If so, it sounds like this is something to be done by the caller not the callee.

daiwaid: > If so, it sounds like this is something to be done by the caller not the callee.

Yes, but in this case since the benchmark dataset processing code is only ever called by the benchmark util code, does it matter which one handles the materialization?

The test set has to be materialized in the callee code because both shuffle and block need it, and I personally think materializing the train set in the caller code makes the code more complex, but I implemented it in #53, if you'd like then merge that PR instead.

zenodflow: I think your Arena module could use benchmark.classic module? At least, down the road, we will have the benchmark module to call non-Libem tools for benchmarking.

zenodflow: >because both shuffle and block need it
if `args.shuffle` is true? And if `args.block` is true? 

Yes, that's indeed the point -- materialize only when it is necessary and for the callers, they are expecting an iterator and will call list() upon the first use of the passed in dataset objects.

zenodflow:
#52

PR #57, opened 2024-07-28
feat: allow match to report likelihood as output

zenodflow:


zenodflow (commented):
Note: the confidence output doesn't go hand in hand with likelihood at the moment. Pending on #55.

PR #58, opened 2024-07-29
feat: redesigned arena api to operate independently of front-end

daiwaid:


zenodflow:
Nice work. Is it possible to add a few integration tests for the server backend? Or a few instructions on how to test the key functionalities of the backend APIs.

zenodflow (approved):

Issue #67, opened 2024-08-01
benchmark/suite unexpected behavior


Char15Xu:
1. When running benchmark/suite/llama3.py, get keyError:name in /libem/libem/prepare/datasets/amazon_google.py.
2. Setting `args.no-schema = True` does not remove the structured format as expected. (currently need to set `args.schema = False` to make it work)

daiwaid:
1. Cannot reproduce. There is no `name` key referenced in /libem/libem/prepare/datasets/amazon_google.py. Can you provide a full error log?
2. Intended behavior, the `--no-schema` flag is mapped to `args.schema'. The flag is named that way because `args.schema` is on by default.

Char15Xu:
1. when running benchmark/suite/llama3.py, the following error happens after finished running abt-buy dataset:
```
Traceback (most recent call last):
  File "/Users/charlesxu/libem/benchmark/suite/llama3.py", line 96, in <module>
    reports = run_benchmark()
  File "/Users/charlesxu/libem/benchmark/suite/llama3.py", line 33, in run_benchmark
    report = run.run(args)
  File "/Users/charlesxu/libem/benchmark/run.py", line 18, in run
    return benchmark_func(args)
  File "/Users/charlesxu/libem/benchmark/classic/amazon_google.py", line 43, in run
    test_set = list(test_set)
  File "/Users/charlesxu/libem/libem/prepare/datasets/amazon_google.py", line 48, in read
    parsed_data['left'] = {field: left_values[field] for field in fields}
  File "/Users/charlesxu/libem/libem/prepare/datasets/amazon_google.py", line 48, in <dictcomp>
    parsed_data['left'] = {field: left_values[field] for field in fields}
KeyError: 'name'
```
2. i think `python benchmark/suite/llama3.py --no-schema` does not remove the structure format. Is setting `args.schema = False` the only way to remove structure format?

daiwaid:
1. There shouldn't be a `name` field in the amazon-google dataset, make sure your libem-sample-data folder is up to date and make sure you aren't adding/modifying the `args.kwargs['fields']` field (try checking if that field exists and what it contains at the beginning of benchmark/classic/amazon_google.py and see if it's been modified to include an additional `name` field, you can see what the `fields` dict is supposed to be set to by default on line 33: `["title", "manufacturer", "price"]`.
[...]

PR #69, opened 2024-08-02
fix: reset args.kwargs before running benchmark suites

Char15Xu:
This pr fixed keyError issue when running different datasets in all files from benchmark/suite.

Relavent issue: #67  bullet point 1

issue: args.kwargs kept values from old dataset.

fIx: reset args.kwargs before running dataset

### DIFF
--- a/benchmark/suite/gpt-35-turbo.py
+++ b/benchmark/suite/gpt-35-turbo.py
@@ -24,7 +24,8 @@ def run_benchmark():
     reports = {}
     for dataset in datasets:
         args.name = dataset
-
+        args.kwargs = {}
+        
         print(f"Running {dataset}...")
         print(args)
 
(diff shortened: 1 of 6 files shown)


zenodflow:
@Char15Xu What is the use of `args.kwargs`? It feels like we need a bigger change for the `args.kwargs` design. It's mysterious why one needs to reset it.

Char15Xu:
Based on my understanding of the original code, it is used in libem/prepare/datasets to parse data from datasets.

zenodflow:
@Char15Xu @daiwaid Instead of relying on the client code to reset the kwargs -- which is really an unknown unknown for them - we should https://github.com/abcsys/libem/blob/b16907cdd5334f915ef585cbf74fc86a2c1db3ff/benchmark/classic/abt_buy.py#L38 do a deep copy of the args here!

daiwaid:
Alternate fix implemented in PR #73.

PR #73, opened 2024-08-02
fix: create a deep copy of args in benchmark

daiwaid:
To avoid modifying a potentially user-defined variable.


zenodflow:
Nice fix.

PR #74, opened 2024-08-03
refactor benchmark/suite and add batch, block suite

daiwaid:


benchmark/classic/abt_buy.py
```diff
@@ -54,6 +55,13 @@ def run(args):
             "libem.match.prompt.experiences": Experiences(),
             "libem.match.prompt.output": ""
         })
-        
+    
+    # set pre-trained similarity cutoff for blocking
+    if args.block:
+        libem.calibrate({
+            "libem.block.parameter.similarity": args.similarity
+            if 0 <= args.similarity <= 100
```
zenodflow, line 63: no need for this condition

zenodflow (approved):
Overall, very nice changes! Left only one minor comment.

zenodflow:
@daiwaid Upon closer look and run, there are several issues that need to be addressed.

First:

> /Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/runpy.py:126: RuntimeWarning: 'benchmark.run' found in sys.modules after import of package 'benchmark', but prior to execution of 'benchmark.run'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))

Also: 

In the benchmark suite, the configurations are more complicated than necessary. E.g., the benchmark suite should always print and log the results.

zenodflow:
The `benchmark/README.md` contains confusing instructions, e.g. `-n gpt-4o` this is very easy to confuse with `-m gpt-4o`. Besides,  https://github.com/abcsys/libem/blob/0b6e38c918cf68458ebe4258796363fd1abbe9a8/Makefile#L90 are not updated to reflect the name changes.

The overall changes are good. Just that they can be simplified. Please see if you can address them in a follow-up PR. (no need to revert).

zenodflow:
Let's address these in #76

PR #76, opened 2024-08-03
refactor benchmark/suite and add batch, block suite

zenodflow:


zenodflow (commented):
Copying from #74 

First:

/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/runpy.py:126: RuntimeWarning: 'benchmark.run' found in sys.modules after import of package 'benchmark', but prior to execution of 'benchmark.run'; this may result in unpredictable behaviour
warn(RuntimeWarning(msg))

Also:

In the benchmark suite, the configurations are more complicated than necessary. E.g., the benchmark suite should always print and log the results.

The `benchmark/README.md` contains confusing instructions, e.g. `-n gpt-4o` this is very easy to confuse with `-m gpt-4o`. Besides,  https://github.com/abcsys/libem/blob/0b6e38c918cf68458ebe4258796363fd1abbe9a8/Makefile#L90 are not updated to reflect the name changes.

Makefile
```diff
@@ -87,21 +87,27 @@ archive:
...
+.PHONY: block-suite batch-suite
+block-suite:
+	python -m benchmark.run -s block
+batch-suite:
+	python -m benchmark.run -s batch
+
 .PHONY: gpt-4o gpt-4o-mini gpt-4 gpt-4-turbo gpt-3.5-turbo
 gpt-4o:
-	python -m benchmark.suite.gpt-4o
+	python -m benchmarkrun -s gpt-4o
```
zenodflow, line 98: typo

Makefile
```diff
@@ -87,21 +87,27 @@ archive:
 	@zip -r "$$(date +%Y-%m-%d_%H-%M-%S)_benchmarks.zip" $(RESULT_DIRS) \
 	&& rm -rf $(RESULT_DIRS) || true
 
+.PHONY: block-suite batch-suite
```
zenodflow, line 90: nice

Makefile
daiwaid: fixed

benchmark/README.md
```diff
@@ -12,12 +12,12 @@ Available benchmarks in the `classic` directory:
...
 - [challenging](https://github.com/abcsys/libem-sample-data/tree/main/challenging)
 
-## Blocking
+## Entity Blocking
 
-To run the blocking benchmarks in `/classic`:
+To run the blocking benchmark:
 
 ```
-python -m benchmark.run --name <benchmark-name> --block --no-match --num-pairs -1
```
zenodflow, line 20: we should keep instruction on single benchmark run.

perhaps say something along the line "To run single blocking benchmark ..."; "to run the entire suite: .."

benchmark/README.md
```diff
@@ -31,25 +31,25 @@ similarity score cutoff between 1-100.
...
 
 \* abt-buy dataset is blocked without the `description` field.
 
 ## Entity Matching
 
-To run the matching benchmarks in `/classic`:
+To run through all the datasets in `/classic`:
 
 ```
-python -m benchmark.run --name <benchmark-name> --num-pairs -1
```
zenodflow, line 52: ditto

benchmark/README.md
```diff
@@ -31,25 +31,25 @@ similarity score cutoff between 1-100.
...
+python -m benchmark.run -s <model-name>
```
zenodflow, line 52: this <model-name> can lead to confusion here. Need to clarify that this is suite name and we provide suites for each model

benchmark/README.md
```diff
@@ -115,14 +128,13 @@ python -m benchmark.run --name <benchmark-name> --num-pairs -1
 
 ### Prompt-level batching
 
-To run the benchmark with batching, use the `--batch-size` flag with the desired batch size. For example, to run
-the `abt-buy` benchmark with a batch size of 16, use the following command:
+To run the batching benchmark for a dataset, use:
 
 ```
-python -m benchmark.run --name abt-buy --num-pairs -1 --batch-size 16
```
zenodflow, line 122: ditto

benchmark/README.md
```diff
@@ -115,14 +128,13 @@ python -m benchmark.run --name <benchmark-name> --num-pairs -1
...
+python -m benchmark.run -s batch -k '{"dataset": <dataset-name>}'
```
zenodflow, line 134: I really don't like the -k here, it's a mess to pass in a dict; 

what about just using `-n`? didn't give it much thought, pls revisit though

benchmark/__init__.py
```diff
@@ -13,24 +13,9 @@
...
-    'walmart-amazon': classic.walmart_amazon.run,
-}
-
-challenges = {
-    'classic.challenging': classic.challenging.run,
-}
+from benchmark.classic import classic_benchmarks
+from benchmark.suite import benchmark_suites
 
 benchmarks = {
```
zenodflow, line 19: if so, can simplify:

L16 ... as benchmarks
L17 ... as suites

benchmark/classic/__init__.py
```diff
@@ -22,3 +22,23 @@
     walmart_amazon,
     challenging,
 )
+
+dataset_benchmarks = {
```
zenodflow, line 26: less clear naming.

keep it as it was.

benchmark/classic/__init__.py
```diff
@@ -22,3 +22,23 @@
...
+    'fodors-zagats': fodors_zagats.run,
+    'itunes-amazon': itunes_amazon.run,
+    'walmart-amazon': walmart_amazon.run,
+}
+
+challenges = {
+    'classic.challenging': challenging.run,
+}
+
+classic_benchmarks = {
```
zenodflow, line 41: Now that you have refactor this dictionary inside the classic package, you don't have to call it `classic benchmarks` but just `benchmarks`, so people may just import it as benchmark.classic.benchmarks

You can do classic_benchmarks, challenges, then benchmarks

benchmark/classic/abt_buy.py
```diff
@@ -54,6 +55,13 @@ def run(args):
             "libem.match.prompt.experiences": Experiences(),
             "libem.match.prompt.output": ""
         })
-        
+    
+    # set pre-trained similarity cutoff for blocking
+    if args.block:
+        libem.calibrate({
+            "libem.block.parameter.similarity": args.similarity
+            if 0 <= args.similarity <= 100
```
zenodflow, line 63: do we really need this error handling condition

benchmark/run.py
```diff
@@ -34,9 +37,6 @@ def run_from_file(args) -> dict:
     "fp", "fn", "tn", "tp" keys or directly in the JSON file.
     """
     
-    # create a deep copy of args before making changes
-    args = copy.deepcopy(args)
```
zenodflow, line 38: why not

benchmark/run.py
```diff
@@ -156,8 +159,47 @@ def args() -> argparse.Namespace:
     return parser.parse_args()
 
 
+def validate(args):
```
zenodflow, line 162: wonderful. will take another look at this.

relevant: #38

benchmark/suite/gpt_4_turbo.py
```diff
@@ -0,0 +1,56 @@
+import os
+import time
+
+from benchmark.classic import dataset_benchmarks
+from benchmark.suite.util import (run_dataset,
```
zenodflow, line 5: Nit: new line after first ( and indent it short. new line at )

benchmark/suite/gpt_4_turbo.py
```diff
@@ -0,0 +1,56 @@
+import os
+import time
+
+from benchmark.classic import dataset_benchmarks
```
zenodflow, line 4: import benchmark as bm

benchmark/suite/gpt_4_turbo.py
```diff
@@ -0,0 +1,56 @@
...
+from benchmark.suite.util import (run_dataset, 
+                                  report_to_dataframe, 
+                                  tabulate, plot, 
+                                  save, show)
+
+name = os.path.basename(__file__).replace(".py", "")
+
+
+def run(args):
+    datasets = dataset_benchmarks.keys()
```
zenodflow, line 14: bm.benchmarks

benchmark/suite/gpt_4_turbo.py
```diff
@@ -0,0 +1,56 @@
...
+    args.block = False
+    args.match = True
+    args.model = "gpt-4-turbo"
+    args.num_pairs = -1
+    
+    # do not use downstream log/print
+    args.log = False
+    args.quiet = True
+    
+    print(f"Benchmark: Matching all {len(datasets)} datasets "
```
zenodflow, line 25: running all { .. } match benchmarks

benchmark/suite/gpt_4_turbo.py
```diff
@@ -0,0 +1,56 @@
...
+        "cost": "Cost ($)",
+        "throughput": "Throughput (pps)",
+    }
+    df = df.rename(columns=field_names)
+    
+    tabulate(df, name)
+    plot(df)
+    show(df)
+
+    return reports
```
zenodflow, line 56: overall, this is much cleaner.

benchmark/suite/gpt_4o.py
```diff
@@ -0,0 +1,56 @@
+import os
+import time
+
+from benchmark.classic import dataset_benchmarks
+from benchmark.suite.util import (run_dataset,
```
zenodflow, line 5: ditto

benchmark/util.py
```diff
@@ -112,27 +111,23 @@ def benchmark(train_set: Iterable,
...
-
     # prepare the blocking dataset
     # deduplicate left and right descriptions
     left = [json.loads(i) for i in set(json.dumps(d['left']) for d in test_set)]
     right = [json.loads(i) for i in set(json.dumps(d['right']) for d in test_set)]
     true = [{'left': d['left'], 'right': d['right']}
             for d in test_set if d['label'] == 1]
 
-    total_pairs = len(left) * len(right)
+    max_pairs = len(left) * len(right)
```
zenodflow, line 121: nice

benchmark/util.py
```diff
@@ -112,27 +111,23 @@ def benchmark(train_set: Iterable,
...
 
-    print(f"Benchmark: Blocking {total_pairs} potential pairs "
+    print(f"Benchmark: Blocking {max_pairs} potential pairs "
           f"from the {args.name} benchmark", end='')
+    
     if args.num_pairs > 0:
+        total_pairs = args.num_pairs + args.start_index
```
zenodflow, line 127: why? I don't get it

zenodflow (commented):
Overall, this is much better. Pls address comments.

benchmark/classic/abt_buy.py
daiwaid: Yes because the default value is -1 (which means replace with calibrated score).

benchmark/run.py
daiwaid: This should be an internal function, all external calls should be routed through run(), so this will just be a duplicate deep copy.

benchmark/util.py
daiwaid: Total pairs needed for the matching phase (skip first `args.start_index` pairs and match `args.num_pairs` pairs). I think I will remove `args.start_index` since it's rarely used and confusing to calculate total pairs.

benchmark/README.md
daiwaid: I think it's more confusing to use `-n` because that is for specifying a benchmark, not a dataset, and the `-k --kwargs` flag is specified as "Additional args that apply to specific benchmark files, in JSON format" so this should be the flag to use.

daiwaid: After further thinking, I agree we should use `-n` because batching can be run on non-dataset-specific benchmarks.

benchmark/README.md
```diff
@@ -12,12 +12,18 @@ Available benchmarks in the `classic` directory:
 - [walmart-amazon](https://github.com/abcsys/libem-sample-data/tree/main/walmart-amazon)
 - [challenging](https://github.com/abcsys/libem-sample-data/tree/main/challenging)
 
-## Blocking
+## Entity Blocking
 
-To run the blocking benchmarks in `/classic`:
+To run blocking through a single benchmark in `/classic`:
```
zenodflow, line 17: To run a single blocking benchmark

benchmark/README.md
```diff
@@ -12,12 +12,18 @@ Available benchmarks in the `classic` directory:
...
 
 ```
-python -m benchmark.run --name <benchmark-name> --block --no-match --num-pairs -1
+python -m benchmark.run -n <benchmark-name> -p -1 --block --no-match
+```
+
+To run through the entire blocking benchmark:
```
zenodflow, line 23: To run all blocking benchmarks:

benchmark/README.md
```diff
@@ -87,10 +99,23 @@ python -m benchmark.run --name <benchmark-name> --num-pairs -1
 | itunes-amazon  |    100    | 46.15  | 63.16 | 0.004256 |        34        |
 | walmart-amazon |   96.3    | 67.36  | 79.27 | 0.03037  |        82        |
 
-### GPT-3.5-turbo
+#### GPT-4
```
zenodflow, line 102: Do we still have the gpt-4-turbo results somewhere?

benchmark/classic/abt_buy.py
zenodflow: "libem.block.parameter.similarity": args.similarity or block_similarities['abt-buy']

benchmark/util.py
```diff
@@ -112,27 +111,24 @@ def benchmark(train_set: Iterable,
...
 
-    print(f"Benchmark: Blocking {total_pairs} potential pairs "
+    print(f"Benchmark: Blocking {max_pairs} potential pairs "
           f"from the {args.name} benchmark", end='')
+    
     if args.num_pairs > 0:
         print(f",\nBenchmark: stopping after the first "
               f"{'pair' if args.num_pairs == 1 else f'{args.num_pairs} pairs'} "
               f"that pass{f'es' if args.num_pairs == 1 else ''} the cutoff:")
+        print("Benchmark: Note that most stats are unavailable when running "
```
zenodflow, line 130: This message makes people worried. Also I don't know what it means / why it is necessary.

benchmark/util.py
```diff
@@ -164,19 +161,26 @@ def run_block(test_set: Iterable, args: argparse.Namespace):
...
-        if pair not in tp:
-            fn.append(pair)
-
-    num_tn = total_pairs - len(tp) - len(fp) - len(fn)
-    if len(tp) == 0:
-        precision = 100 if len(fp) == 0 else 0
-        recall = 100 if len(fn) == 0 else 0
+    # if didn't run through the entire dataset
+    # then all F1 related stats are not available
+    if args.num_pairs > 0 and args.num_pairs <= len(out):
```
zenodflow, line 166: if 0 < args.num_pairs <= len(out)

benchmark/util.py
```diff
@@ -164,19 +161,26 @@ def run_block(test_set: Iterable, args: argparse.Namespace):
...
+        fn = []
+        precision, recall, f1 = -1, -1, -1
     else:
-        precision = len(tp) / (len(tp) + len(fp)) * 100
-        recall = len(tp) / (len(tp) + len(fn)) * 100
-    f1 = 0 if precision + recall == 0 \
-        else 2 * round(precision * recall / (precision + recall), 2)
+        for pair in true:
+            if pair not in tp:
+                fn.append(pair)
```
zenodflow, line 173: fn = [ ... ] use list comprehension to slice it

benchmark/util.py
```diff
@@ -190,22 +194,30 @@ def run_block(test_set: Iterable, args: argparse.Namespace):
...
 
-    print()
     print(f"Benchmark: Blocking done in {round(total_time, 2)}s.")
-    if args.num_pairs <= 0 or args.num_pairs > len(out):
-        print(f"Benchmark: Resulting pairs\t {len(out)}")
-        print(f"Benchmark: Percent blocked\t {round((1 - len(out) / total_pairs) * 100, 2)}")
-    print(f"Benchmark: Precision\t\t {stats['precision']}")
-    print(f"Benchmark: Recall\t\t {stats['recall']}")
-    print(f"Benchmark: F1 score\t\t {stats['f1']}")
+    if not args.quiet:
```
zenodflow, line 213: combine the two if statements.

benchmark/util.py
```diff
@@ -190,22 +194,30 @@ def run_block(test_set: Iterable, args: argparse.Namespace):
             'fn': len(fn)
         }
     }
-
+    
+    if args.num_pairs <= 0 or args.num_pairs > len(out):
```
zenodflow, line 198: could args.num_pairs be equal to len(out)

benchmark/util.py
daiwaid: It is what it says: it's impossible to calculate any stats related to F1 without running through the entire dataset so most of the stats will be -1 (since it requires a lot of extra work to retrieve the pairs filtered out by libem.block()).

benchmark/util.py
zenodflow: why args.num_pairs could be == 0. I am confused by this condition

zenodflow (commented):
Looks better. Left some more comments. 

@daiwaid

benchmark/util.py
daiwaid: `args.num_pairs` will always (uness `args.num_pairs > total pairs`) equal to `len(out)` since we stop when `len(out) == args.num_pairs`.

benchmark/util.py
daiwaid: Actually, I will implement the code to retrieve the filtered out pairs, and we can decide if it's worth it to keep or not.

zenodflow: I don't know why this message helps me.

zenodflow: Could we just keep the original logic and messages we had? What is the intended effect with the new changes?

benchmark/util.py
zenodflow: I wonder why args.num_pairs could be 0

This if clause is very confusing.

benchmark/util.py
daiwaid: The original logic was wrong (miscalculating tn and fn pairs) so this message was put in place. I will implement a way to get the right tn and fn pairs.

benchmark/util.py
daiwaid: when `args.num_pairs` <= 0 it means run through the entire benchmark.

daiwaid:
Addressed all comments. Take a look @zenodflow.

benchmark/README.md
daiwaid: Yes but it was not ran with the latest benchmark tools so the cost field is not calculable (only have total tokens but not input vs output tokens).

benchmark/util.py
zenodflow: ok

benchmark/util.py
zenodflow: including == 0?

benchmark/README.md
```diff
@@ -31,25 +37,31 @@ similarity score cutoff between 1-100.
...
-| dblp-acm      |   678927    |        79         |        99.9        |   50.3   |
-| dblp-scholar  |   823244    |        59         |        99.2        |   7.32   |
-| fodors-zagats |    13224    |        83         |        99.8        |   80.0   |
-| itunes-amazon |    7488     |        63         |        95.8        |   15.2   |
-| walmart-amazon|   716846    |        50         |        96.7        |   1.6    |
+|   Benchmark   | Total Pairs | Similarity Cutoff (0-100) | Percentage Blocked |    F1    | Throughput (pps) |
+| :------------ | :---------: | :-----------------------: | :----------------: | :------: | :--------------: |
+| abt-buy*      |   367136    |            50             |        95.6        |   2.54   |      86000       |
+| amazon-google |   460106    |            54             |        96.4        |   2.8    |      79000       |
+| beer          |    6308     |            79             |        98.8        |   30.4   |      59000       |
```
zenodflow, line 44: If the tp and fp calculations were off, should the F1 be different?

benchmark/suite/util.py
```diff
@@ -0,0 +1,58 @@
+import copy
+import os
+import pandas as pd
+from datetime import datetime
+
```
zenodflow, line 5: insert a new line

benchmark/suite/util.py
```diff
@@ -0,0 +1,58 @@
...
+import libem
+import benchmark as bm
+
```
zenodflow, line 8: insert a new line

benchmark/suite/util.py
```diff
@@ -0,0 +1,58 @@
...
+def run_benchmark(benchmark, args):
+    # create a deep copy of args to pass into benchmark
+    args_cpy = copy.deepcopy(args)
+
+    libem.reset()
+    return bm.benchmarks[benchmark](args_cpy)
```
zenodflow, line 14: just 

libem.reset()
return bm.benchmarks[benchmark](copy.deepcopy(args))

benchmark/README.md
daiwaid: The wrong calculations only affected runs that did not go through entire datasets (i.e. had `--num-pairs` flag set to > 0) but the suite always ran through entire datasets so these results are unaffected.

benchmark/suite/util.py
daiwaid: Done.

benchmark/README.md
zenodflow: Ack. We can add a new pass in future

zenodflow (commented):
lgtm

PR #81, opened 2024-08-09
feat: add duckdb and mongodb integration to cluster

daiwaid:


libem/resolve/cluster/integrations/duckdb.py
```diff
@@ -0,0 +1,95 @@
+import re
+import duckdb
```
zenodflow, line 2: Update requirements.txt

libem/resolve/cluster/integrations/duckdb.py
```diff
@@ -0,0 +1,95 @@
...
+import pandas as pd
```
zenodflow, line 3: Dynamic import

libem/resolve/cluster/integrations/duckdb.py
zenodflow: May need dynamic import too

libem/resolve/cluster/integrations/duckdb.py
```diff
@@ -0,0 +1,95 @@
...
+    clusters = cluster_func(
+            df.to_dict(orient="records")
+        )
+    
+    df["__cluster__"] = [cluster_id for cluster_id, _ in clusters]
+    
+    if sort:
+        df = df.sort_values(by="__cluster__")
+    
+    return input.add(df)
```
zenodflow, line 95: This module has clean code. Liked it. Nice work.

I'll revisit in the next pass.

libem/resolve/cluster/integrations/mongodb.py
```diff
@@ -0,0 +1,80 @@
+import re
+import pymongo.database
```
zenodflow, line 2: update requirements.txt

dynamic import

libem/resolve/cluster/interface.py
```diff
@@ -20,20 +24,30 @@
     list[(ID, str)],
     list[(ID, dict)],
     "pd.DataFrame",
+    str
 ]
 
 
 def cluster(records: InputType, *, sort=False) -> OutputType:
     import pandas as pd
+    from libem.resolve.cluster.integrations.duckdb import DuckDBTable
```
zenodflow, line 33: I see, if this is the first import of the two modules, no need to do dynamic import previously. Pls double-check with `make time_mod_load` in case mod time increases.

zenodflow (changes requested):
Nice work. See if you can add an `examples/apps/duckdb.py` and an `examples/apps/mongodb.py`.

libem/resolve/cluster/interface.py
```diff
@@ -20,20 +24,30 @@
...
+            return mongo_func(records, sort)
         case pd.DataFrame():
             from libem.resolve.cluster.integrations.pandas import func as pandas_func
-
-            if sort:
-                return pandas_func(records).sort_values(by="__cluster__")
-            else:
-                return pandas_func(records)
+            
+            return pandas_func(records, sort)
```
zenodflow, line 50: nice

examples/apps/cluster.py
```diff
@@ -18,10 +18,10 @@ def main():
     df_test = df.drop(["cluster_id"], axis=1)
 
     with libem.trace as t:
-        df_cluster = libem.cluster(df_test)
+        df_cluster = libem.cluster(df_test, sort=True)
```
zenodflow, line 21: You cannot do this here. The evaluation would be wrong.

libem/resolve/cluster/function.py
```diff
@@ -60,18 +62,14 @@ def func(records: Iterable[Record]) -> list[(ClusterID, Record)]:
...
-    }
+    # reassign cluster ids to records to increment from 0 without gaps
+    for i, cluster in enumerate(cluster_id_records.keys()):
+        for record_digest in cluster_id_records[cluster]:
+            record_cluster_ids[record_digest] = i
+    
     return [
-        (new_cluster_id_map[cluster_id], record)
-        for record, cluster_id in zip(records, record_cluster_ids.values())
+        (record_cluster_ids[digest(record)], record)
```
zenodflow, line 71: this is an additional digest call

libem/resolve/cluster/function.py
```diff
@@ -60,18 +62,14 @@ def func(records: Iterable[Record]) -> list[(ClusterID, Record)]:
...
-    # remap cluster ids to increment from 0 without gaps
-    unique_cluster_ids = sorted(
-        id for id in set(record_cluster_ids.values())
-    )
-    new_cluster_id_map = {
-        old_id: new_id
-        for new_id, old_id in enumerate(unique_cluster_ids)
-    }
+    # reassign cluster ids to records to increment from 0 without gaps
+    for i, cluster in enumerate(cluster_id_records.keys()):
```
zenodflow, line 66: The cluster_id_records will contain clusters without any records. Not sure if this is correct.

Makefile
```diff
@@ -146,3 +146,10 @@ dedupe:
 	python examples/apps/dedupe.py
 link:
 	python examples/apps/link.py
+
+# libem integrations
+.PHONY: duckdb mongodb
+duckdb:
+	python examples/integrations/duckdb-integration.py
```
zenodflow, line 153: just duckdb.py and mongodb.py

examples/apps/cluster.py
zenodflow: In fact, the default expected behavior of cluster is simply assigning __cluster__ without sorting. This is intended as the example.

examples/integrations/duckdb-integration.py
```diff
@@ -0,0 +1,31 @@
...
+import libem
+from libem.prepare.datasets.clustering import febrl
+from libem.resolve.cluster.integrations.duckdb import DuckDBTable
+
+def main():
+    random.seed(1)
+    num_samples = 10
+    
+    # load in dataset and remove cluster_id
+    df = febrl.load_test() \
```
zenodflow, line 13: create a new method that load data to mongo.

examples/integrations/mongodb-integration.py
```diff
@@ -0,0 +1,67 @@
+import pprint
+import random
+import pymongo
+
+import libem
+from libem.prepare.datasets.clustering import febrl
+from libem.resolve.cluster.integrations.mongodb import MongoCollection
+
+def get_available_name(basename, names) -> str:
```
zenodflow, line 9: put this to later

examples/integrations/mongodb-integration.py
```diff
@@ -0,0 +1,67 @@
...
+    
+        return f"{basename}_{index}"
+    return basename
+
+def main():
+    random.seed(1)
+    num_samples = 5
+    
+    # load in dataset and remove cluster_id
+    df = febrl.load_test() \
```
zenodflow, line 26: ditto, create a method that loads

libem/resolve/cluster/function.py
daiwaid: There will not be any empty clusters since `pop()` removes the key along with the values.

Makefile
daiwaid: That doesn't work, the `import duckdb` line from libem/resolve/cluster/integrations/duckdb.py will try to import this file rather than the actual library.

libem/resolve/cluster/function.py
daiwaid: This is the only way (I can think of) to be able to retain duplicate records.

daiwaid:
Addressed comments. Force pushed to remove changes to `libem/resolve/cluster/function.py` and move them to its own PR.

libem/resolve/cluster/integrations/duckdb.py
zenodflow: FYI: @daiwaid I checked out the code. Although the implementation looks clean, the design could be simplified a lot. I'll do a pass.

zenodflow (approved):

PR #85, opened 2024-08-09
refactor: simplify cluster id assignment

daiwaid:


libem/resolve/cluster/function.py
```diff
@@ -33,21 +33,20 @@ def func(records: Iterable[Record]) -> list[(ClusterID, Record)]:
...
-            else:
-                cluster_id_records[cluster_id].extend(
-                    cluster_id_records.pop(left_cluster_id)
-                )
-            for record_digest in cluster_id_records[cluster_id]:
-                record_cluster_ids[record_digest] = cluster_id
+            
+            # merge the cluster with the bigger id 
+            # into the smaller id cluster when matched
+            small_id_cluster = min(left_cluster_id, right_cluster_id)
```
zenodflow, line 42: Keep the original

libem/resolve/cluster/function.py
```diff
@@ -60,18 +59,14 @@ def func(records: Iterable[Record]) -> list[(ClusterID, Record)]:
...
-    # add cluster id to records
-    # remap cluster ids to increment from 0 without gaps
-    unique_cluster_ids = sorted(
-        id for id in set(record_cluster_ids.values())
-    )
-    new_cluster_id_map = {
-        old_id: new_id
-        for new_id, old_id in enumerate(unique_cluster_ids)
-    }
+    # reassign cluster ids to records to increment from 0 without gaps
```
zenodflow, line 62: @daiwaid Does this achieve the same as the original version

daiwaid: Yes

zenodflow: Meaning, the original goal was to assign the lowest cluster_id to 0, second lowest to 1, and so on.

This code does not guarantee so, although it actually achieves so by relying on the insertion order of cluster_id -- however, this is implicit / hard to tell from the code.


Nit:
```
    # reassign cluster ids so that they increment from 0 without gaps
```

daiwaid: Yes, the insertion order guarantees this so we don't need the extra code (could add a comment to make it explicit), and this new approach actually improves on the original if the goal is to have the cluster_id order follow the record list order for the case with duplicate records:

Assume the dataset only has 1 column and looks like: `['dog', 'cat', 'dog']`, then line 24 will look like `{'cat': 1, 'dog': 2}` because the duplicate record will overwrite the original cluster_id, so at the end, the original version will assign 'cat' a 0 and both 'dog' records a 1, even though 'dog' appeared earlier in the list. This new version addresses this because it only relies on the insertion order and not the old cluster_id order.

zenodflow: I see. Thanks for the explanation. But does this apply to the new code after #82 was in?

libem/resolve/cluster/function.py
zenodflow: @daiwaid I can see where you come from. Either version is fine.

 For this one, consider smaller / bigger --> lower / higher. I.e., small_id_cluster -> low_cluster_id etc.

libem/resolve/cluster/function.py
daiwaid: Yes, the code in #82 still experiences this edge case, but this PR will not.

daiwaid:
Addressed all changes, ready for merge.

libem/resolve/cluster/function.py
```diff
@@ -61,18 +60,15 @@ def func(records: Iterable[Record]) -> list[(ClusterID, Record)]:
...
+    # the final cluster_id order will follow the input records order
+    for i, cluster in enumerate(cluster_id_records.keys()):
+        for record_digest in cluster_id_records[cluster]:
+            record_cluster_ids[record_digest] = i
+    
     return [
-        (remap_cluster_id[record_cluster_ids[d]], r)
-        for r, d in zip(records, digests)
+        (record_cluster_ids[digest], record)
+        for digest, record in zip(digests, records)
```
zenodflow, line 71: the name `digest` collides with outside method name, make it `d` and `r`

libem/resolve/cluster/function.py
```diff
@@ -61,18 +60,15 @@ def func(records: Iterable[Record]) -> list[(ClusterID, Record)]:
...
-    unique_cluster_ids = sorted(
-        id for id in set(record_cluster_ids.values())
-    )
-    remap_cluster_id = {
-        old_id: new_id
-        for new_id, old_id in enumerate(unique_cluster_ids)
-    }
+    # reassign cluster ids so that they increment from 0 without gaps, 
+    # the final cluster_id order will follow the input records order
+    for i, cluster in enumerate(cluster_id_records.keys()):
```
zenodflow, line 65: cluster -> cluster_id?

libem/resolve/cluster/function.py
```diff
@@ -34,21 +34,20 @@ def func(records: Iterable[Record]) -> list[(ClusterID, Record)]:
...
-            else:
-                cluster_id_records[cluster_id].extend(
-                    cluster_id_records.pop(left_cluster_id)
-                )
-            for record_digest in cluster_id_records[cluster_id]:
-                record_cluster_ids[record_digest] = cluster_id
+            
+            # merge the cluster with the highger id 
+            # into the lower id cluster when matched
+            low_id_cluster = min(left_cluster_id, right_cluster_id)
```
zenodflow, line 43: low_id_cluster -> low_cluster_id

zenodflow:
Thanks. Some more naming nits.

daiwaid:
Done.

PR #94, opened 2024-08-26
feat: add plotting to benchmark suites

daiwaid:
- Add plotting to benchmark suites; use `--plot` flag to plot from the last benchmark suite run results.

Addresses #

```release-note
- Add --plot flag to the benchmark module to plot the last benchmark suite run
- New benchmark suite runs will automatically save plots to 
```

### DIFF
--- a/benchmark/suite/gpt_35_turbo.py
+++ b/benchmark/suite/gpt_35_turbo.py
@@ -8,10 +8,21 @@ from benchmark.suite.util import (
     tabulate, plot, save, show
 )
 
-name = os.path.basename(__file__).replace(".py", "")
+name = os.path.basename(__file__).replace(".py", "").replace('_', '-').replace('35', '3.5')
 
 
 def run(args):
+    if args.plot:
+        files = list(os.scandir(bm.result_dir))
+        for file in reversed(files):
+            if file.name[26:-4].replace('35', '3.5') == name:
+                print(f"Plotting run: {file.name}")
+                plot(file)
+                return
+        
+        print(f"Could not find past {name} suite results to plot.")
+        return
+    
     benchmarks = bm.classic_benchmarks.keys()
 
     args.block = False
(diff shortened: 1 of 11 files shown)


benchmark/suite/gpt_35_turbo.py
```diff
@@ -8,10 +8,21 @@
...
 
-name = os.path.basename(__file__).replace(".py", "")
+name = os.path.basename(__file__).replace(".py", "").replace('_', '-').replace('35', '3.5')
 
 
 def run(args):
+    if args.plot:
+        files = list(os.scandir(bm.result_dir))
+        for file in reversed(files):
+            if file.name[26:-4].replace('35', '3.5') == name:
```
zenodflow, line 18: What is 26:-4?

benchmark/suite/gpt_35_turbo.py
```diff
@@ -35,7 +46,7 @@ def run(args):
     print(f"Benchmark: Suite done in: {time.time() - start:.2f}s.")
 
     df = report_to_dataframe(reports)
-    save(df, name)
+    file_name = save(df, name)
```
zenodflow, line 49: The logic here could be simplified by introducing a load call in the if args.plot?

zenodflow (changes requested):
Nice work - just wonder if the logic can be simplified a bit.

benchmark/suite/gpt_35_turbo.py
daiwaid: to extract the suite name from the result file (remove date and .csv).

benchmark/suite/gpt_35_turbo.py
daiwaid: The issue with that is we still need to pass the .csv file name into `plot()` so it can save the plot with the same name as the result. We can't generate the file name like here because in the `if args.plot` part, we are using an old result so the timestamp will be wrong if we just pass in the DataFrame.

daiwaid:
Improved version in PR #95.

PR #96, opened 2024-09-03
fix: add option to change LIBEM_SAMPLE_DATA_PATH in config.yaml

daiwaid:
When the libem module is installed through pip, `LIBEM_SAMPLE_DATA_PATH` will not point to the correct/desired location, so the option to change the path is needed.

Addresses #

```release-note
NONE
```


zenodflow:
This is good fix. Could you also add a line or two in the benchmark/README.md to inform readers whether/what/how to configure? I.e., letting them know that they should use `libem configure` had they installed via `pip`.

zenodflow (approved):

zenodflow (approved):

PR #101, opened 2024-09-30
feat: add ability to reset libem.calibrate calls

daiwaid:
- add ability to reset libem.calibrate calls (necessary for libem serve functionality)

Addresses #

```release-note
- libem.reset() no longer resets external model clients, use libem.model.reset() instead
```

### DIFF
--- a/libem/tune/calibrate/function.py
+++ b/libem/tune/calibrate/function.py
@@ -10,6 +10,7 @@ from libem.core.struct import (
 schema = {}
 
 
+# Cache the original/non-calibrated values to allow for resetting
 cache = {}
 
 def func(config, verbose=False):
--- a/libem/tune/calibrate/function.py
+++ b/libem/tune/calibrate/function.py
@@ -10,6 +10,8 @@ from libem.core.struct import (
 schema = {}
 
 
+cache = {}
+
 def func(config, verbose=False):
     for path, value in config.items():
         # Split the path into module and attribute parts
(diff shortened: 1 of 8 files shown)


libem/tune/calibrate/function.py
```diff
@@ -24,13 +26,22 @@ def func(config, verbose=False):
         except:
             raise Exception(f"unable to find parameter {path}")
         assert isinstance(parameter, Parameter), f"invalid parameter {path}"
+        
+        # Cache the original value
+        if path not in cache:
+            cache[path] = parameter()
```
zenodflow, line 32: Liked this simple change! But was wondering if we should use the parameter struct itself to store cache value vs. this "external" cache.

zenodflow: I prefer the current design - but let's have a think over the alternative?

daiwaid: I think it depends on what we define as "reset". Currently the design is to reset `libem.calibrate()` calls (modifications directly from the user), so other means of modifying parameters (i.e. optimizing via grid search, loading from a config file, etc) will not be reset. If we want to strictly reset all parameters to their default values then moving this to the Parameter class may be better.

zenodflow: Got it. That makes sense!

libem/tune/calibrate/function.py
```diff
@@ -24,13 +26,22 @@ def func(config, verbose=False):
         except:
             raise Exception(f"unable to find parameter {path}")
         assert isinstance(parameter, Parameter), f"invalid parameter {path}"
+        
+        # Cache the original value
+        if path not in cache:
```
zenodflow, line 31: Does this support multiple reset calls? Or just reset to the first / original value?

libem/tune/calibrate/function.py
```diff
@@ -10,6 +10,8 @@
 schema = {}
 
 
+cache = {}
```
zenodflow, line 13: Could we add a brief explanation of how the cache works?

libem/tune/calibrate/function.py
daiwaid: It resets to the original (non-calibrated) value.

libem/tune/calibrate/function.py
daiwaid: Done. It only saves the parameter value the first time said parameter is being calibrated so it only saves the original/non-calibrated value.

PR #102, opened 2024-10-13
feat: rework libem serve with support for oauth

daiwaid:
- Reworked libem serve with OAuth
- Accessible at https://serve.libem.org

Addresses #

```release-note
NONE
```


requirements.txt
```diff
@@ -14,10 +14,6 @@ numpy
 pandas
 tabulate
 bs4
-flask
```
zenodflow, line 17: Just to be sure - are there any modules outside serve/ rely on these dependencies?

Btw., we should consider separate the libem / arena / serve/ benchmark dependencies with individual requirements.txt.

serve/libem.yaml
```diff
@@ -14,18 +14,23 @@ spec:
     spec:
       containers:
       - name: libem-web-service
-        image: silveryfu/libem-serve:latest
+        image: daiwaid/libem-serve:latest
```
zenodflow, line 17: Make sure the repo is public on docker registry

serve/requirements.txt
```diff
@@ -0,0 +1,9 @@
+uvicorn
+requests
+fastapi
+itsdangerous
+python-dotenv
+fastapi-login
+authlib
+mysql-connector-python
+https://github.com/abcsys/libem/archive/refs/heads/main.zip
```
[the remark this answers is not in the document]
zenodflow: Btw., I see you already do the requirements separation here! Nice.

In future, we should reflect this in the main Makefile, e.g, `make install-serve`, `make install-arena` `make install-benchmark` etc.

zenodflow (commented):
Great update. I wonder if it is possible to add a simple, local e2e test case for the serve? E.g., start the serve and return a "guess" result.

Also, perhaps copy over the README in arena to here (tailor it a bit for serve)?

serve/requirements.txt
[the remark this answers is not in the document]
daiwaid: This is to add the latest version of libem package without needing to use git.

requirements.txt
daiwaid: Yeah I added individual requirements.txt for arena and serve and removed those packages from the main file. It's all packages used for API so it shouldn't be used anywhere else.

serve/libem.yaml
daiwaid: Should we make a libem account?

requirements.txt
zenodflow: Great!

zenodflow (approved):

PR #104, opened 2024-10-22
add libem compete cli command

daiwaid:
- `libem compete` cli command demonstrates the interaction between Libem Arena and Libem Serve

Addresses #

```release-note
NONE
```

### DIFF
--- a/cli/libem
+++ b/cli/libem
@@ -5,6 +5,7 @@ import yaml
 import ast
 import argparse
 import pprint
+import requests
 
 import libem
 import libem.prepare.datasets as datasets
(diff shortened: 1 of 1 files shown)


cli/libem
```diff
@@ -5,6 +5,7 @@ import yaml
 import ast
 import argparse
 import pprint
+import requests
```
zenodflow, line 8: This should be in the requirements.txt

zenodflow:
How about calling it `libem live`? As in, live in action, live demo, etc.

I get what `compete` implies but it's a bit lengthy and have somewhere departed connotations. 

More comments to come! Nice work.

@daiwaid

zenodflow (approved):

PR #105, opened 2024-10-27
feat: add pvc storage and mysql to serve

daiwaid:
- add pvc storage and mysql to serve
- support "offline mode" without google authentication by setting `OFFLINE_MODE` environment variable to `true`

Addresses #

```release-note
NONE
```


serve/serve.env
```diff
@@ -0,0 +1,7 @@
+OPENAI_API_KEY=""
+CLIENT_ID=""
+CLIENT_SECRET=""
+TRIAL_CREDITS=5000
+SECRET=secret
+SQL_PASSWORD=password
```
zenodflow, line 6: In our current deployment, we have a more sophisticated password, right? :)

daiwaid: Yes.

PR #106, opened 2024-10-28
Bug fix for pricing issue

Char15Xu:
This PR updates `model_info` when the model choice changes during matching. This ensures that the correct `model_info` is used for retrieving pricing.

Addresses #  

```release-note
```

### DIFF
--- a/libem/optimize/cost/openai.py
+++ b/libem/optimize/cost/openai.py
@@ -1,6 +1,7 @@
 from libem.optimize.cost import cache
 
 model_info = None
+model_choice = None
 null_model_info = {
     'input_cost_per_token': 0,
     'output_cost_per_token': 0
@@ -18,8 +19,9 @@ def get_input_cost(model, num_tokens):
     if num_tokens is None or num_tokens <= 0:
         return 0
 
-    global model_info
-    if model_info is None:
+    global model_info, model_choice
+    if model_info is None or model_choice != model:
+        model_choice = model
         model_info = get_model_info(model)
     return model_info['input_cost_per_token'] * num_tokens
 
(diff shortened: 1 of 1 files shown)


zenodflow:
Good fix @Char15Xu. How do we calculate the cost for anthropic models?

zenodflow:
@daiwaid Do you think this bug impact our existing results (e.g., the graphs on blog)?

daiwaid:
> @daiwaid Do you think this bug impact our existing results (e.g., the graphs on blog)?

No, all benchmark tests are done by explicitly passing in the currently used model into the `get_cost()` function.

Char15Xu:
> Good fix @Char15Xu. How do we calculate the cost for anthropic models?

The same code is used for anthropic models

zenodflow:
> > Good fix @Char15Xu. How do we calculate the cost for anthropic models?
> 
> The same code is used for anthropic models

I see. @Char15Xu but do we have the cost cache for the anthropic models?

Char15Xu:
> > > Good fix @Char15Xu. How do we calculate the cost for anthropic models?
> > 
> > 
> > The same code is used for anthropic models
> 
> I see. @Char15Xu but do we have the cost cache for the anthropic models?

Sorry, I missed this message. When doing benchmark, anthropic models also use the libem.optimize.cost.openai module. There is not a separate cost cache for anthropic models.

zenodflow:
> > > > Good fix @Char15Xu. How do we calculate the cost for anthropic models?
> > > 
> > > 
> > > The same code is used for anthropic models
> > 
> > 
> > I see. @Char15Xu but do we have the cost cache for the anthropic models?
> 
> Sorry, I missed this message. When doing benchmark, anthropic models also use the libem.optimize.cost.openai module. There is not a separate cost cache for anthropic models.

Should there be one? If so, we should do a future PR for this.

zenodflow (approved):

PR #108, opened 2024-11-21
fix: improve confidence score generation

daiwaid:
- move the confidence score to be generated after the answer and simplify the prompt
- import prompts from `libem.core.struct.pattern` to `libem.match.prompt` so these prompts are saved as part of benchmark config dump

Addresses #

```release-note
```

### DIFF
--- a/libem/match/function.py
+++ b/libem/match/function.py
@@ -295,32 +295,50 @@ def parse_output(output: str) -> dict:
     """
     # remove empty lines and process lines in reverse order
     output = [s for s in output.splitlines() if s][::-1]
-
-    answer, confidence, explanation = "no", None, None
-
-    if parameter.confidence():
-        for i, line in enumerate(output):
-            line = line.lower()
-            nums = re.findall(r"\d+\.\d+|\d+", line)
-            if nums:
-                confidence = float(''.join(nums))
-                output = output[i + 1:]
-                break
     
-    for i, line in enumerate(output):
-        line = line.lower()
-        nums = re.findall(r"\d+\.\d+|\d+", line)
-        if prompt.output.value == Index("likelihood") and nums:
-            answer = float(answer)
-            output = output[i + 1:]
-            break
-        elif 'yes' in line or 'no' in line:
-            answer = "yes" if "yes" in line else "no"
-            output = output[i + 1:]
-            break
+    index = 0
+    answer, confidence, explanation = None, None, None
+    
+    # parse for numeric values (confidence, likelihood)
+    # with priority given to likelihood if 
+    # both are enabled but only 1 value is found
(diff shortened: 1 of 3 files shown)


zenodflow:
@daiwaid nice! Is there evidence showing the before vs. after results?

E.g., we need some form of calibration evaluation here.

daiwaid:
250 pairs from `amazon-google` using `gpt-4o-2024-08-06`:

Before:
output

After:


Even though the model is in general overconfident in the new results, the wrong answers have lower confidence scores as expected, whereas before, the wrong answers had high confidence scores. In addition, there's about a 15% token saving due to the shortening of the prompt.

zenodflow:
I see. Lgtm

libem/match/function.py
```diff
@@ -312,6 +306,18 @@ def parse_output(output: str) -> dict:
                 confidence = float(''.join(nums))
                 output = output[i + 1:]
                 break
+    
+    for i, line in enumerate(output):
```
zenodflow, line 310: @daiwaid Could we merge this for loop with the for loop above?

libem/match/prompt.py
```diff
@@ -2,6 +2,9 @@
 from libem.core.struct.prompt import (
     Shot, Rules, Experiences
 )
+from libem.core.struct.pattern import (
+    CoT, Confidence
```
zenodflow, line 6: Is this used/referenced anywhere?

zenodflow: Ah I see, from the function.py. Good change.

daiwaid:
I revised the parsing logic to account for more outlier cases (when confidence and answer are given on the same line). The code is a bit complicated but this is as simple as I can make it. I included comments so hopefully it's clear.

daiwaid:
The new implementation considers more edge cases that were previously not considered:
1. if the answer/likelihood and the confidence score are given on the same line
2. if the model doesn't give a confidence score, only a likelihood score

Since this parsing logic applies to all models, we can't guarantee that all models will format things the way `gpt-4o` does. Another option is to specify the formatting in the prompt (some kind of guardrail), but that will increase the cost. We can't do structured output since not all models support it.

We can also choose to not deal with these edge cases since they are rare (with gpt models), in that case some slight modifications are still necessary to avoid potentially running into exceptions when calling `output.pop()` when there are no more lines left.

zenodflow:
I see. Thanks. The code path should perhaps split for ones that support structured output (e.g., gpt) and those do not so that the brittleness is contained. Let me read your code again and see if we can simplify / make clearer.

zenodflow:
#109 is now in.

zenodflow (approved):

PR #109, opened 2024-12-01
feat: support structured output for the latest OpenAI models

daiwaid:
- support structured output for the latest OpenAI models
- add `libem.match.parameter.likelihood` parameter to toggle likelihood output
- add `likelihood` to benchmark
- slightly refactor `core/model/openai.py`

Addresses #

```release-note
action required:
- use parameter `libem.match.parameter.likelihood` to toggle likelihood output
- with `likelihood` enabled, `answer` field from `match()` calls will still be yes/no, likelihood output will be in the `likelihood` field
- NOTE: DuckDuckGo search may raise rate limit error: https://github.com/deedy5/duckduckgo_search/issues/215
```

### DIFF
--- a/examples/tool/browse.py
+++ b/examples/tool/browse.py
@@ -8,6 +8,7 @@ def positive():
     print("No browse:")
     libem.calibrate({
         "libem.match.parameter.tools": [],
+        "libem.match.parameter.model": "gpt-4o"
     })
 
     is_match = libem.match(e1, e2)
--- a/examples/tool/browse.py
+++ b/examples/tool/browse.py
@@ -7,8 +7,7 @@ def positive():
 
     print("No browse:")
     libem.calibrate({
-        "libem.match.parameter.tools": [],
-        "libem.match.parameter.model": "gpt-4o"
+        "libem.match.parameter.tools": []
     })
 
     is_match = libem.match(e1, e2)
--- a/examples/tool/likelihood.py
+++ b/examples/tool/likelihood.py
@@ -9,20 +9,20 @@ def main():
     is_match = libem.match(left, right)
 
     print("Entity 1:", left, "\nEntity 2:", right)
-    print("Match likelihood:", is_match["answer"])
+    print("Match likelihood:", is_match["likelihood"])
 
 
     left, right = "apple inc.", "Applebee's"
     is_match = libem.match(left, right)
 
     print("Entity 1:", left, "\nEntity 2:", right)
-    print("Match likelihood:", is_match["answer"])
+    print("Match likelihood:", is_match["likelihood"])
 
     left, right = "apple", "Apple electronics"
     is_match = libem.match(left, right)
 
     print("Entity 1:", left, "\nEntity 2:", right)
-    print("Match likelihood:", is_match["answer"])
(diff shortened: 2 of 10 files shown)


libem/match/prompt.py
```diff
@@ -21,26 +25,25 @@
...
+                            "are the same.")
+    else:
+        output_string.append("Give your answer in the form of a "
+                            "single 'yes' or 'no'.")
+    if output_type() == "strict":
+        output_string.append("Nothing else.")
+    
+    return ' '.join(output_string)
+    
+output = build_prompt
```
zenodflow, line 46: This changes the original logic, right?

examples/tool/browse.py
```diff
@@ -8,6 +8,7 @@ def positive():
     print("No browse:")
     libem.calibrate({
         "libem.match.parameter.tools": [],
+        "libem.match.parameter.model": "gpt-4o"
```
zenodflow, line 11: Is there a reason why we have to specify the model?

examples/tool/likelihood.py
```diff
@@ -1,11 +1,8 @@
 import libem
 
-from libem.core.struct.parameter import Index
-
-
 def main():
     libem.calibrate({
-        "libem.match.prompt.output": Index("likelihood")
+        "libem.match.parameter.likelihood": True
```
zenodflow, line 5: Is there a reason why we make likelihood a standalone option?

libem/core/model/__init__.py
```diff
@@ -4,6 +4,9 @@
 from libem.core import exec
 import libem
 
+def structured_schema(*args, **kwargs) -> dict:
```
zenodflow, line 7: Should it be better named as output_schema?

zenodflow: Also, is there a reason why this is exposed for the model module?

libem/match/prompt.py
daiwaid: No.

examples/tool/browse.py
daiwaid: > NOTE: the browse example no longer consistently trigger the browse tool when using structured outputs

libem/core/model/__init__.py
daiwaid: This can be used for more than just output schema, it can also create tool/function schema, etc.

It is used in `libem.match`: https://github.com/abcsys/libem/blob/95e6f00285b45933eeda53259dfa12ca9e9c6bcd/libem/match/function.py#L171

examples/tool/browse.py
zenodflow: Do we know why this happens? Is it because of the use of structured output or something else?

zenodflow: Do we know why this happens? Is it because of the use of structured output or something else?

examples/tool/likelihood.py
zenodflow: My take was it's because it is part of the `output` configuration (as in, match output)

daiwaid:
Updated top comment with new changes.

PR #110, opened 2024-12-02
Adding cascade feature

Char15Xu:
This PR introduces the cascade functionality to Libem, enabling the system to dynamically adjust model usage based on confidence scores. This aims to optimize cost and throughput while maintaining the accuracy of GPT-4o.

Key changes include:
- `vectorize`, `prematch`, and `match` under Libem/cascade.  
  - Add reset functionality to clear the client before `prematch` and `match` stages.  
- Under `Libem/examples/cascade`:  
  - `util.py`: Contains benchmark and plot functions to measure cost and accuracy.  
  - `online.py`: Provides a working example of the cascade functionality.  
  - `bench.py`: Performs benchmarking and generates graphs to visualize cost and accuracy.  

Addresses # NA

#### Special notes for your reviewer:
I am only adding the reset as part of the cascade since it is the most reasonable place I can think of for now. Would love to improve on this later if there is a better place to do reset. 

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -172,3 +172,13 @@ cython_debug/
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 .idea/
 
+# Ignore macOS .DS_Store files
+.DS_Store
+examples/.DS_Store
+examples/cascade/.DS_Store
+
+# Ignore specific output folder
+examples/cascade/output/
+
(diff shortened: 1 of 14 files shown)


zenodflow:
Nice. Will review this coming week.

zenodflow:
Nice - will review and work through this / next week.

PR #111, opened 2025-01-21
feat: implement record batching with structured output

daiwaid:
- implement smart batching by grouping shared records to reduce input tokens
- implement batching with structured output to reduce chance of hallucination

Addresses #

```release-note
NONE
```


libem/core/model/openai.py
```diff
@@ -38,16 +38,16 @@ def get_client():
     return _client
 
 
-def output_schema(name, **fields) -> dict:
+def output_schema(_name, **fields) -> dict:
```
zenodflow, line 41: Nit: why underscore?

daiwaid: if one of the fields in the schema is called "name", then this will prevent the error of passing in two parameters with the same name.

zenodflow (changes requested):
* Run through 1-2 benchmarks to see if accuracy changes
* Make smart batching a tunable parameter with default on

daiwaid:
Running through the smart_batch example:
```
F1 difference: 0.04
Speedup: 0.99x
Cost savings: $0.029 (46.0%)
```

Note: smart batching generally matches normal batching performance but can perform better/worse for specific pairings.

zenodflow:
> Speedup: 0.99x

Is there a reason why the performance is the same?

Makefile
```diff
@@ -58,6 +58,8 @@ async_request:
 	python examples/optimize/async_request.py
 async_batch:
 	python examples/optimize/async_batch.py
+smart_batch:
```
zenodflow, line 61: I think `smart_batch` is a bit vague as a name and may infer a mixed, high-level strategy that picks among prompt sharing, async execution, and sharing left record pairs. Perhaps `record_batch`?

libem/match/parameter.py
```diff
@@ -72,6 +72,10 @@
     default=1
 )
 
+smart_batch = Parameter(
```
zenodflow, line 75: `record_batch`

Btw., is the accuracy issue now stabilized?  Is it safe to enable it be default?

zenodflow:
Overall this is great but have two minor comments

libem/match/parameter.py
daiwaid: it has improved from slightly changing the prompt, it still performs worse than regular batching sometimes but not catastrophically. I did not run any large-scale tests.

daiwaid:
> > Speedup: 0.99x
> 
> Is there a reason why the performance is the same?

The latency mostly depends on the number of output tokens, while smart batch decreases input tokens, it has the same # of output tokens so latency should be roughly equal (there is also a lot of variance to the OpenAI API latency as well).

libem/match/parameter.py
zenodflow: I see. Let's make it default off then?

daiwaid:
Implemented requested changes @zenodflow .

zenodflow (approved):

PR #112, opened 2025-01-28
multiprocessing support for blocking

daiwaid:
- add multiprocessing to blocking to speed up for large datasets
- add prepare for the full amazon google dataset

Addresses #

```release-note
- add multiprocessing to blocking to speed up for large datasets
- block no longer returns a generator and will always return the full result as a list
```


zenodflow (approved):
For future:
* Use Ray instead of multiprocess
* Use Ray Data for dataset partitioning

PR #113, opened 2025-02-25
feat: multimodal support for match

daiwaid:
- support any combination of text and/or image input for match

Addresses #

```release-note
```


test/match.py
```diff
@@ -1,21 +1,80 @@
+import os
```
zenodflow, line 1: This test is great. Let's also add an example (one without assertion) using the same pics.

libem/match/struct.py
```diff
@@ -0,0 +1,160 @@
...
+# input types
+TextEntityField = str
+TextEntityFields = TextEntityField | dict[str, TextEntityField]
+ImageEntityField = Image.Image | np.ndarray
+ImageEntityFields = ImageEntityField | dict[str, ImageEntityField]
+
+
+class MultimodalEntityDesc(TypedDict):
+    text_fields: TextEntityFields | None
+    image_fields: ImageEntityFields | list[ImageEntityField] | None
```
zenodflow, line 18: Is the idea here allowing multiple images (e.g., consecutive ones from a video clip) to be placed in the desc?

daiwaid: yes, each entity can have multiple images (for different angles, interior vs exterior, etc)

libem/match/struct.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+EntityDesc = TextEntityFields | list[TextEntityFields] | \
+             ImageEntityFields | list[ImageEntityFields] | \
+             MultimodalEntityDesc | list[MultimodalEntityDesc]
```
zenodflow, line 23: Nice. An alternative is to have the TextEntity and ImageEntity inherit the MultimodalEntity - which allows adding built-in methods in each class level to do entity specific processing without invoking methods in the more generic desc type. But the opposite - which is the current way - composing the latter from the former two is better and can achieve the same thing at the caller. We just have to keep in mind that at the caller there will be multiple client code (as what's in prompt.py) and we _should have_ perhaps add a multiplexer to hide the complexity from match.py -- though fortunately, context window of LLMs can contain any form of data without us doing that extra step :)

zenodflow (commented):
LGTM on first look. Do we have green light on the original "test cases" (i.e. the examples)?

libem/match/struct.py
zenodflow: Just to clarify, my point is just - whether it is deliberate in your design - it is because context window of LLMs can just work with any type of data, the composition design makes more sense (otherwise the internal code gonna mess up a bit - right now we just have to handle them in prompt.py).

That said, how are we handling the image on the input side? By url or bytes or both?

daiwaid:
Added multimodal example.

> LGTM on first look. Do we have green light on the original "test cases" (i.e. the examples)?

Yes, all examples run as expected.

libem/match/struct.py
daiwaid: The reason for the composition design is to allow for primitive types as valid inputs (e.g. str) in addition to the custom defined dict types. Also the current implementation do not create actual classes but just using the TypedDict class to create type hints.

Both image types are now supported.

zenodflow:
Lgtm.

zenodflow (approved):

benchmark/util.py
```diff
@@ -331,7 +331,7 @@ def run_match(train_set, test_set, args):
                 continue
 
             match = span['match']
-            left, right = match['left'], match['right']
+            left, right = match['left']['text'], match['right']['text']
```
zenodflow, line 334: Minor (for improvement): When the second selector, 'text' is not given, could we have a defaulter that returns the text.

zenodflow:
@daiwaid Let's consider it in follow-up PR but I wonder if we can improve the struct naming a bit:
* EntityDesc -> Record (i.e. MultimodalRecord). A record is a description - whether it matches to an entity is the application level business. Record should be the core struct and internal representation. It is a basic data type in other data systems such as Zed - so later we will be easily covert between the two.
Let me know what you think!

PR #114, opened 2025-03-04
feat: support blocking across n datasets

daiwaid:
- redesign block module
  - support operation across n datasets
  - use Ray instead of multiprocessing for parallel computation
  - add support for multimodal input (same as match module)
- standardize input types/structs
  - make input typing hints/structs from the match module reusable (now in `libem.struct`)
  - rename EntityDesc -> Record
  - create digest function to support all valid input types
- add n-way input support for dedupe and link
  - currently only supports pandas and dict types

Addresses #

```release-note
- remove `left` and `right` parameters from `libem.block()`, now support passing in any number of datasets
```

### DIFF
--- a/examples/match.py
+++ b/examples/match.py
@@ -3,7 +3,7 @@ import cv2
 import libem
 
 from libem.match.prompt import rules
-from libem.match.struct import MultimodalEntityDesc
+from libem.struct import MultimodalRecord
 
 
 parent_directory = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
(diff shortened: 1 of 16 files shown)


zenodflow:
> use Ray instead of multiprocessing for parallel computation

Just to confirm before merging: this change will come in follow-up PR or this one?

daiwaid:
> > use Ray instead of multiprocessing for parallel computation
> 
> Just to confirm before merging: this change will come in follow-up PR or this one?

This one, it's already implemented.

zenodflow:
> > > use Ray instead of multiprocessing for parallel computation
> > 
> > 
> > Just to confirm before merging: this change will come in follow-up PR or this one?
> 
> This one, it's already implemented.

I see it now - the block/function.py was collapsed in the file review panel. Just to confirm, the `import ray` is used only in the block.py, right?

daiwaid:
> > > > use Ray instead of multiprocessing for parallel computation
> > > 
> > > 
> > > Just to confirm before merging: this change will come in follow-up PR or this one?
> > 
> > 
> > This one, it's already implemented.
> 
> I see it now - the block/function.py was collapsed in the file review panel. Just to confirm, the `import ray` is used only in the block.py, right?

Yes.

